PR #24, opened 2018-01-09
rewrite using chained implicits instead of macros for 0.3.0

erikerlandson:
The macro-based implementation of `0.2.0` has a couple unappealing properties.

Firstly, one of the quirks of resolving implicit values inside a macro context is that it can only resolve them if they are part of a completely compiled unit. So you cannot declare a new unit and use it in the same compilation.

Secondly, lack of implicit parameters in the `Quantity` method signatures makes it difficult or impossible to integrate 0.2.0 with any other package (refined, pureconfig, etc).  It also forced me to specifically implement logic for various spire types, which means it is not possible to add new types just by defining the appropriate spire implicit values, which of course somewhat defeats the benefits of type classes.

This 0.3.0 PR is a total re-write of the underlying logic to use no macros, only chained implicits and dependent types. It should be relatively straightforward to integrate with other typelevel packages and allows one to define new units and use them immediately.


erikerlandson:
Currently this rewrite is being thwarted by Scala's aggressive cycle checking during implicit resolution. As a possible fix I am proposing https://github.com/scala/scala/pull/6383

erikerlandson:
There is also a more general fix from Miles Sabin, slated for 2.13: https://github.com/scala/scala/pull/6050
This may mean future releases of coulomb will only work for 2.13+

erikerlandson:
I'm not even sorry.
https://github.com/scala/scala/pull/6438

erikerlandson:
Scala 2.13.0-M5 has the Miles Sabin fix above. I think the one missing dep I have is a spire build for 2.13.0-M5 (https://github.com/non/spire/issues/742)

PR #93, opened 2020-07-16
Use Typeable instead of reflect

cquiroz:
This PR is mostly to open a discussion as I'm not totally sure it is the correct solution.

To enable scala.js we need to remove the use of `scala.reflect`. This PR replaces this by shapeless' `Typeable`. This PR runs all the tests correctly and I tested this on scala.js and it works fine

### DIFF
--- a/coulomb/src/main/scala/coulomb/infra/canonical.scala
+++ b/coulomb/src/main/scala/coulomb/infra/canonical.scala
@@ -47,21 +47,7 @@ object NoImplicit {
 }
 
 object TypeString {
-  import scala.reflect.runtime.universe._
-  def typeString[T :WeakTypeTag]: String = {
-    def work(t: Type): String = {
-      t match { case TypeRef(pre, sym, args) =>
-        val ss = sym.toString
-                    .stripPrefix("free ")
-                    .stripPrefix("trait ")
-                    .stripPrefix("class ")
-                    .stripPrefix("type ")
-        val as = args.map(work)
-        if (args.length <= 0) ss else (ss + "[" + as.mkString(",") + "]")
-      }
-    }
-    work(weakTypeOf[T])
-  }
+  def typeString[T :Typeable]: String = Typeable[T].describe
 }
 
 trait IsUnitExpr[T] {
@@ -91,13 +77,12 @@ trait GetBaseUnit[U] {
   def bu: BaseUnit[U]
 }
 trait GetBaseUnitP1 {
-  import scala.reflect.runtime.universe._
 
   implicit def undeclared[T, TUE](implicit
       enabled: coulomb.policy.EnableUndeclaredBaseUnits,
       testUE: IsUnitExpr.Aux[T, TUE],
       notUE: TUE =:!= true,
-      tt: WeakTypeTag[T]): GetBaseUnit[T] = {
+      tt: Typeable[T]): GetBaseUnit[T] = {
     val name = TypeString.typeString[T]
     new GetBaseUnit[T] {
       val bu = new BaseUnit[T](name, name)
--- a/.gitignore
+++ b/.gitignore
@@ -15,3 +15,6 @@ project/plugins/project/
 # Scala-IDE specific
 .scala_dependencies
 .worksheet
+.bloop/
+.hydra/
+.metals/
--- a/coulomb/src/main/scala/coulomb/define/define.scala
+++ b/coulomb/src/main/scala/coulomb/define/define.scala
@@ -17,7 +17,6 @@ limitations under the License.
 package coulomb.define
 
 import scala.language.implicitConversions
-import scala.reflect.runtime.universe._
 
 import spire.math._
 import shapeless._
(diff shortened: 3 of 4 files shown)


coulomb/src/main/scala/coulomb/infra/canonical.scala
```diff
@@ -47,21 +47,7 @@ object NoImplicit {
 }
 
 object TypeString {
-  import scala.reflect.runtime.universe._
```
cquiroz, line 50: The change looks like an oversimplification but there isn't that much info you can get from a `Typeable[T]`

erikerlandson:
Thanks, this looks promising!  I'm interested in how this interacts with `QuantityParser`.  `withImports` might become mandatory to compensate for not having fully qualified type names.  Sort of a shame that `Typeable` doesn't include fully qualified name.

Alternatively, could just not support QuantityParser for scala.js, but it precludes some of the other integrations.

erikerlandson:
What is the effect of functions having `implicit wtt: WeakTypeTag[T]` in the signature on scala.js compilation, if they are not used?

If I added a scala.js specific methods to `QuantityParser` that used `Typeable`, and required imports, but kept the existing methods as-is for use in scala, does scala.js handle that as long as one only uses the scala-js-specific methods in scala.js?

cquiroz:
When you add `WeakTypeTags` surpisingly the scala.js side compiles but when you build an actual js it goes into the linking phase where it will attempt to locale the scala.js version of each of the classes on the dependency tree and fail to link, which is what I got when trying to compile the tests to scala.js

cquiroz:
You can provide though jvm and js specific versions so it maybe possible to keep using `WeakTypeTag` on jvm and `Typeable` on scala.js in some cases. This could be done by having two versions on `QuantityParser` or extract the different parts into a platform specific trait

erikerlandson:
If we leverage the `jvm` and `js` versions, I am inclined to define a shim `TypeName[T]`
```scala
trait TypeName[T] {
  def name: String // fully qualified name
}
object TypeName {
  implicit def manifestTypeName[T]: TypeName[T] =
    // on JVM, wraps `implicit WeakTypeTag`
    // on JS, invokes macro, similar to below
}
```

And the macro call would be implemented similar to `typeName_impl` 
[here](https://stackoverflow.com/a/15649874/3669757):
```scala
def typeName[T]: String = macro typeName_impl[T]

def typeName_impl[T](c: Context): c.Expr[String] = {
  import c.universe._

  val TypeApply(_, List(typeTree)) = c.macroApplication
  c.literal(typeTree.toString())
}
```

Then the code divergence between js and jvm should be limited to the definition of `TypeName[T]`, and the macro-paradise dependency is only needed for scala.js

Everywhere else in the code would (I think) just replace WeakTypeTag[T] with TypeName[T].
I haven't tested that this works, but seems plausible

cquiroz:
That sounds like an interesting approach, seems plausible and I may try it

cquiroz:
I'm trying to get this to work and the `TypeName[T]` style works in jvm using `WeakTypeTag` but I'm having troubles with the macro. I have tried several styles and I always get as the type name `T` as (I suppose) begin erased
I'm not able to find so much info about macros. Do you have any pointers

erikerlandson:
@cquiroz  the macros I've used to properly bind types have involved `WeakTypeTag`, for example:
https://github.com/erikerlandson/coulomb/blob/v0.2.0/macros/src/main/scala/com/manyangled/coulomb/UnitMacros.scala#L786

the situation with `shapeless.Typeable` seems more ambiguous but it also makes use of `WeakTypeTag`
https://github.com/milessabin/shapeless/blob/master/core/src/main/scala/shapeless/typeable.scala#L391

Possibly if WeakTypeTag is used only in the context of these macro bindings, it sidesteps the problems with scala.js, since the actual invocation is just a type and it is only the macro that "sees" WeakTypeTag?

cquiroz:
I have never done a macro before so all my opinions are highly suspect. You can use WeakTypeTags in macros (even in js) but it seems the type info gets lost, I often get `T` as the type name. I'll keep looking

erikerlandson:
Another approach to try is to build the macro using scala only, as a separate sub-package, that the other packages depend on. I did this in coulomb-0.2.0, back when I was using macros: all my macros were built first, as a subpackage that the "main" packages depended on.  I don't know if that will work better with scala.js, but it might.

erikerlandson:
Did a bit of experimenting with `ClassTag` which scala.js supports (IIUC).

```scala
scala> implicitly[WeakTypeTag[Meter]].tpe.typeSymbol.fullName
val res118: String = coulomb.si.Meter

scala> implicitly[ClassTag[Meter]].runtimeClass.getCanonicalName
val res126: String = coulomb.si.package.Meter
```
`ClassTag` does not support type parameters, so for example my unit test with using type name `Array[Int]` as a unit would not work in scala.js  However I believe it can be made to correctly support all the normal unit types, which have no type parameters. Actual cases with type parameters like using `Array[Int]` as a unit are interesting to me theoretically but I doubt they are important for practical use cases.

You can also see that it puts `.package.` in the canonical name, which might need to be removed, but would be pretty easy to detect and get rid of.

So in this solution concept, `TypeName` would wrap `WeakTypeTag` for normal jvm, and would wrap `ClassTag` for scala.js, and the difference in support between the two would not be very large.  Most people will probably not even notice.
```scala
trait TypeName[T] {
  def name: String // fully qualified name
}
object TypeName {
  implicit def manifestTypeName[T]: TypeName[T] =
    // on JVM, wraps `implicit WeakTypeTag`
    // on JS, wraps `implicit ClassTag`, (with possible manipulation to make it behave like WeakTypeTag)
}
```

cquiroz:
I got something that seems to work, I'll try to send a PR in a few days and we can discuss alternatives

cquiroz:
Superseded by #97

Issue #65, opened 2020-03-08
Should the unit type `U` in Quantity[N, U] be covariant?


erikerlandson:
```scala
trait Fruit
trait Apple extends Fruit
trait Pear extends Fruit

val q1 = 1.withUnit[Apple] + 1.withUnit[Pear]  // is q1 Quantity[Int, Fruit] ?

def f1(f: Quantity[Int, Fruit]) = { ... }

f1(1.withUnit[Apple])
f1(1.withUnit[Pear])
```

would likely depend on #22

erikerlandson:
Although this is interesting, my attempts so far to make it work have failed, and it would be a major increase in implementation complexity, and also increase difficulty to reason about coulom's behavior, so I'm closing this.

PR #95, opened 2020-07-17
Add cats Eq/Order for Quantity

cquiroz:
This is a first attempt at a `coulomb-cats` module. It contains now just the definitions of `Eq` and `Order` for `Quantity[A, U]`. Hopefully it can be expanded to get us `Functor` but I had some troubles on my attempts

It contains two modules: `coulomb-cats` contains the typeclasses and `coulomb-testkit` that contains `Arbitrary` and `Cogen` instances for test and `discipline` tests. If you prefer I could move the tests to `coulomb-tests`


coulomb-testkit/src/main/scala/coulomb/arb/ArbQuantity.scala
```diff
@@ -0,0 +1,19 @@
+package coulomb.arb
```
erikerlandson: I agree it makes sense for this to be a separate library.  I would call the library `coulomb-scalacheck` and the package name `coulomb.scalacheck`

coulomb-testkit/src/main/scala/coulomb/arb/ArbQuantity.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import org.scalacheck.Arbitrary
+import org.scalacheck.Arbitrary._
+import org.scalacheck.Cogen
+import coulomb._
+
+trait ArbQuantity {
+  implicit def arbQuantity[A: Arbitrary, B]: Arbitrary[Quantity[A, B]] = Arbitrary {
```
erikerlandson: suggest using `V` and `U` instead of `A` and `B`

coulomb-testkit/src/test/scala/coulomb/cats/CatsSpec.scala
```diff
@@ -0,0 +1,18 @@
...
+
+import cats.tests.CatsSuite
+import cats.{ Eq, Show, Order }
+import cats.kernel.laws.discipline._
+import coulomb._
+import coulomb.cats.implicits._
+import coulomb.si._
+import coulomb.siprefix._
+
+final class CatsSpec extends CatsSuite {
```
erikerlandson: maybe move the actual property/unit tests into `coulomb-tests` ?

coulomb-cats/src/main/scala/coulomb/cats/package.scala
```diff
@@ -0,0 +1,24 @@
...
+    * @group typeclass
+    */
+  implicit def eqQuantity[A: Eq, B]: Eq[Quantity[A, B]] = Eq.by(_.value)
+
+  /**
+    * Order[Quantity[A, B]] when there is an Order[A]
+    *
+    * @group typeclass
+    */
+  implicit def orderQuantity[A: Order, B]: Order[Quantity[A, B]] = Order.by(_.value)
```
erikerlandson: So I take it that the unit type `B` is assumed the same in these comparisons?   `Quantity[Float, Meter] < Quantity[Float, Meter]` is safe to compare just by extracting `value` but isn't correct if the units are different

cquiroz: Hmm, that's right. I guess this will only work for example if you're sorting a list with the same unit. Since `B` is `invariant` that should be ok right?

erikerlandson: I think it is OK as long as it's enforced that both LHS and RHS are same value type and unit type. Currently Quantity is invariant.  I have considered what it might mean to make it covariant (#65) but that is very speculative.

This is the main reason I never defined scala `Ordering` - supporting convertable units (e.g. comparing a quantity of meters with a quantity of yards) doesn't really fit in the scheme of "normal" ordering-related typeclasses.

erikerlandson:
This looks good.  One last nit is that the file name `.../cats/package.scala` is typically convention for the file defining a package-object.  I'd name this file `cats/implicits.scala`

coulomb-cats/src/main/scala/coulomb/cats/package.scala
```diff
@@ -0,0 +1,24 @@
+package coulomb.cats
+
+import cats._
+import cats.implicits._
+import coulomb.Quantity
+
+trait CatsImplicits {
+  /**
+    * Eq[Quantity[A, B]] when there is an Eq[A]
```
erikerlandson: type names `V` and `U`

erikerlandson:
Oh, can you add a file `coulomb-cats/README.md`?   An example would be `coulomb-refined/README.md` - some information on how to include in project, maybe a couple short code examples

cquiroz:
Just now that I'm writing the documentation I realize that  `Quantity` already has a `===` operator. It makes me wonder if we really need `Eq` instances, though there are other uses of course like when using scalacheck

erikerlandson:
I think there are potential use cases.  I wanted "unit aware" comparisons but something like sorting an array is likely to be a case of comparing a collection of quantities having all the same value and unit types.

cquiroz:
I added a readme and a few more test cases.
I move the package from `coloumb.cats.implicits` to `coloumb.implicits`. This is because often you import `coloumb._` and the `cats` package would shadow the root `cats` package. Let me know if that seems ok

erikerlandson:
> This is because often you import coloumb._ and the cats package would shadow the root cats package.

That is a very interesting point.  I have to assume this comes up for all of my integrations, as I have `coulomb.refined`, `coulomb.pureconfig`, etc.  Currently I think this would be resolved via something like `import _root_.pureconfig...`, for example [here](https://github.com/erikerlandson/coulomb/tree/develop/coulomb-pureconfig#how-to-import).

In the immediate situation, what I propose is that we stick with `coulomb.cats.implicits`, and I suppose library users would then need `import _root_.cats.implicits._`  which is a bit of a hack but at least it is consistent with the rest of the coulomb integration packages.

In the longer term (a future coulomb-0.6 ?), this makes me want to re-think the package paths I am declaring, which would of course break some backward compatibility, but maybe some convention like `import coulomb.integrations.cats...` or perhaps invert it, like `import cats.coulomb....`

cquiroz:
Ok it makes sense. It is a problem mostly if you try to import something from cats after coulomb

cquiroz:
I'll update it

cquiroz:
I moved the package back as suggested

coulomb-cats/README.md
```diff
@@ -0,0 +1,129 @@
...
+```
+Same units can be compared as expected
+```scala
+@ a.eqv(b)
+res7: Boolean = true
+```
+
+Convertible units can be compared too 
+```scala
+@ @ a.eqv(c)
```
erikerlandson: is this a duplicate ammonite prompt?

cquiroz: Yes, let me clean this up

erikerlandson:
LGTM, Thanks @cquiroz!

PR #97, opened 2020-07-22
Attempt to support scala.js

cquiroz:
This is my attempt at adding support for scala.js in coulomb.

I'll add some comments on the code but some general notes:
* The build has been restructured to do cross compilation where it make sense. Some projects like pureconfig only make sense on the jvm side
* The tests have been divided trying to keep as much as possible running on both jvm and js but some can only run on the jvm
* Serialization tests were moved to its own class
* as coulomb was using `WeakTypeTag` to get type info, the proposed idea of a typeclass in #93 was implemented. The js side is a macro and it seems to work.


build.sbt
```diff
@@ -26,55 +26,64 @@ def docDepSettings = Seq(
   previewSite := {}
 )
 
-lazy val coulomb = (project in file("coulomb"))
+lazy val coulomb = crossProject(JVMPlatform, JSPlatform)
+  .crossType(CrossType.Full)
```
cquiroz, line 31: Most projects can be fully cross built, but `coulomb` needs platform specific implementations of `UnitTypeName`

build.sbt
```diff
@@ -85,8 +94,7 @@ def coulombParserDeps = Seq(
 )
 
 lazy val coulomb_parser = (project in file("coulomb-parser"))
```
cquiroz, line 97: Some projects remain jvm only

coulomb-tests/jvm/src/test/scala/coulomb/SerializationTest.scala
```diff
@@ -0,0 +1,35 @@
...
+import singleton.ops._
+
+import utest._
+
+import coulomb.si._
+import coulomb.temp._
+
+import coulomb.validators.CoulombValidators._
+
+object SerializationTests extends TestSuite {
```
cquiroz, line 17: These tests were taken for `QuantityTest` and `TemperatureTest` as these need to run on the jvm only

coulomb-tests/shared/src/test/scala/coulomb/Quantity.scala
```diff
@@ -330,7 +330,7 @@ object QuantityTests extends TestSuite {
         1.withUnit[Meter].show == ("1 m"),
         1.withUnit[Kilo %* Meter].show == ("1 km"),
         (1.5.withUnit[Meter] / 1.0.withUnit[Second]).show == ("1.5 m/s"),
-        1.0.withUnit[Second].pow[-1].show == ("1.0 s^(-1)"),
+        1.0.withUnit[Second].pow[-1].show == ("1.0 s^(-1)") || 1.0.withUnit[Second].pow[-1].show == ("1 s^(-1)"),
```
cquiroz, line 333: scala.js mentions on the docs that a float `1.0` is printed as `1` without the trailing 0 unlike the jvm. Eventually `show` could take a formatter string but that maybe better dealt separately

coulomb-tests/shared/src/test/scala/coulomb/Quantity.scala
```diff
@@ -520,7 +512,8 @@ object QuantityTests extends TestSuite {
       val q1 = (1.withUnit[MooUnit]) + (1.withUnit[Kilo %* MooUnit])
       assert(q1.show == "1001 MooUnit")
       val q2 = (10.withUnit[Seq[Int]]) / (5.withUnit[Second])
-      assert(q2.show == "2 Seq[Int]/s")
+      // js is not yet able to find the nested type
+      assert((q2.show == "2 Seq[Int]/s") || (q2.show == "2 Seq[A]/s"))
```
cquiroz, line 516: The macro can't yet read the recursive types, hopefully it can be done eventually

coulomb-tests/shared/src/test/scala/coulomb/validators.scala
```diff
@@ -29,18 +28,18 @@ object CoulombValidators {
   }
 
   implicit class WithQuantityShouldMethods[V, U](q: Quantity[V, U])(implicit
-      ttV: WeakTypeTag[V],
-      ttU: WeakTypeTag[U],
+      ttV: UnitTypeName[V],
```
cquiroz, line 31: we use `UnitTypeName` as a local type class to abstract reading the name of the unit

coulomb/js/src/main/scala/coulomb/typename.scala
```diff
@@ -0,0 +1,54 @@
...
+}
+
+object UnitTypeName {
+  @inline
+  def apply[T](implicit ev: UnitTypeName[T]): UnitTypeName[T] = ev
+
+  /**
+    * Build an instance of `UnitTypeName` for the type `A`
+    */
+  def unitTypeNameImpl[A: c.WeakTypeTag](
```
cquiroz, line 19: This macro is partially able to create a new instance of `UnitName` reading the type name from the AST

coulomb/jvm/src/main/scala/coulomb/typename.scala
```diff
@@ -0,0 +1,39 @@
...
+  def =:=[U](t: UnitTypeName[U]): Boolean
+  def wtt: Type // only for internal use, shouldn't be called from client code
+}
+
+object UnitTypeName {
+  @inline
+  def apply[T](implicit ev: UnitTypeName[T]): UnitTypeName[T] = ev
+
+  // Implicit builder for `UnitTypeName` instances
+  implicit def unitTypeName[T](implicit ut: WeakTypeTag[T]): UnitTypeName[T] =
```
cquiroz, line 15: This is more or less ported directly from the code in `develop`

cquiroz (commented):
Added some general comments in the code

build.sbt
erikerlandson: Do you have any thoughts on why coulomb-parser isn't working with scala.js?  My prediction was that if you could get fully-qualified unit type names, then QuantityParser should work

coulomb-tests/shared/src/test/scala/coulomb/Quantity.scala
erikerlandson: This is annoying, but I'm OK with leaving it this way since the number of tests affected won't be very large

build.sbt
cquiroz: QuantityParser calls the reflect api and I didn't want to dig further as I was mostly trying to get the core and units to run.

Also the QuantityParser tests were dependent on pureconfig which is not available in scala.js but I think one test could be extracted to run in both

coulomb/js/src/main/scala/coulomb/typename.scala
erikerlandson: What is your opinion on doing it with the macro, versus using `ClassTag`?   I'm fine with using a macro, mostly just curious.

coulomb/js/src/main/scala/coulomb/typename.scala
```diff
@@ -0,0 +1,54 @@
...
+
+import scala.language.experimental.macros
+
+import scala.reflect.macros.whitebox
+
+import scala.reflect.macros.blackbox.Context
+
+trait UnitTypeName[T] {
+  def name: String // typeName
+  def typeString: String
```
erikerlandson, line 9: is there a reason for having both `name` and `typeString` ?

coulomb-tests/jvm/src/test/scala/coulomb/SerializationTest.scala
erikerlandson: does serialization not apply in JS, or does it not work?

build.sbt
erikerlandson: possibly it doesn't have a lot of application in scala JS.  I don't have a strong feel for it.  If it's easy to make work, I guess I'm in favor of doing it, but OTOH if it is hard then we can leave it for some other time.

erikerlandson:
Overall this LGMT.  The two main questions I'm thinking about are:
1. relative merits of using `ClassTag` versus the macro for UnitTypeName on the JS side
2. Is it worth getting `coulomb-parser` onto JS at this time.

cquiroz:
I could try to get the `parser` to work. About the merits is a of the macro it is a bit hard for me to say. I fear the cost of macro maintenance especially thinking of dotty, and also it maybe that we are generating a lot more code which would affect js size at the end.
Was `ClassTag` deprecated at some moment?

coulomb/js/src/main/scala/coulomb/typename.scala
cquiroz: Yes, the implementation on the jvm side using `WeakTagTypes` differs notably and I wanted to preserve that.

coulomb/js/src/main/scala/coulomb/typename.scala
cquiroz: Basically I did the macro as it was suggested initially. Maybe I can try to do it with `ClassTag` and if all tests pass call it done? It was nice to learn a bit about macros but maybe it is a problem long term. It may induce the creation of a lot of code too but I'm not sure about that

coulomb-tests/shared/src/test/scala/coulomb/Quantity.scala
cquiroz: As a reference:
http://www.scala-js.org/doc/semantics.html

coulomb-tests/jvm/src/test/scala/coulomb/SerializationTest.scala
cquiroz: Serialization isn't supported in scala.js at all

erikerlandson:
Regarding the eventual impacts of scala-3, both implicits and macros involve major changes, so I don't think it will matter a lot which solution we have here: supporting scala3 will be close to a total rewrite for `coulomb`.

Some functions of `ClassTag` are deprecated or no longer supported, however the basic ability of ClassTag to identify fully qualified type names seems to still work.  Its naming of type path names seems to be a bit out of sync with WeakTypeTag, so maybe the macro is better for keeping the JVM and JS sides equivalent.

cquiroz:
I'm trying with `ClassTag` and scala.js support is partial, e.g. `getCanonicalName` is not available. `getSimpleName` works except for custom units. `getName` may work but it needs the name to be parsed

erikerlandson:
> I'm trying with ClassTag and scala.js support is partial, e.g. getCanonicalName is not available. 

Yeah, then let's just keep your macro

cquiroz:
One extra data point. I suspected that using a macro could lead to a larger binary size (in js) as we are generating code in all call sites (I think). The test suite in js with class tag is 9.2m while with macros is 9.8m

erikerlandson:
Did you get a ClassTag implementation that also passes unit tests?

cquiroz:
`ClassTag` can cover most cases but not all. We may be able to get `MooUnit` with some work but `Seq[A]` seems impossible. at most you'd get `Seq`

erikerlandson:
I'm fine to stick with the macro, it sounds like it has better type coverage, and the overall impact is small

erikerlandson:
with respect to code size, removing `typeName`, `=:=`, `wtt`, etc from UnitTypeName will make macro-generated code smaller.  The `name` method is the only necessary one.

cquiroz:
I couldn't fine a way to replace `=:=` in `CoulombValidators` that's why I kept it in the typeclass 

Anyway in terms of size test code maybe not the most representative as we intentionally invoke the macros lots of times.

cquiroz:
This maybe a showstopper to port `quantity-parser`
https://github.com/erikerlandson/coulomb/blob/b7e63da16a6aa794d58530527b29d660748478b6/coulomb-parser/src/main/scala/coulomb/parser/QuantityParser.scala#L48

erikerlandson:
@cquiroz if the runtime mirror and toolbox stuff doesn't have good support on JS, then we should definitely drop coulomb-parser for this pull request.  I've been considering a redesign of the parsing anyway, maybe something aligned with https://github.com/apache/avro/pull/841 that doesn't need to invoke runtime compilation.

cquiroz:
In general reflection is not supported in scala.js, except some very limited uses like getting a class name

erikerlandson:
@cquiroz now that we have answers on my two questions, I think this is ready for final review, do you agree?

cquiroz:
I do agree, it would be great to get a release with scala.js soonish

erikerlandson:
+1, I think we can have this merged and a new release published by sometime this weekend.

coulomb/js/src/main/scala/coulomb/typename.scala
```diff
@@ -0,0 +1,54 @@
+package coulomb
+
+import scala.language.experimental.macros
+
+import scala.reflect.macros.whitebox
```
erikerlandson: are both blackbox and whitebox used here?

erikerlandson:
I noticed you moved some files into `.../shared/...`  but others did not move (for example `coulomb-si-units` and the other unit packages).  Mostly just curious why some had to move and others did not.  Is it necessary to use `shared` when there are other non-shared files?

coulomb/js/src/main/scala/coulomb/typename.scala
cquiroz: With my basic understanding of macros I think whitebox are preferred but I could try if the test work in either case
I think only one should be used

cquiroz:
The dirs depend if `CrossType` is set to `Pure`, meaning there is no platform specific code and all code stay in `src`. Or it can be `Full` in which case you have `shared`, `jvm` and `js` dirs 

I prefer to be `Pure` whenever possible but in two cases we need full. In `coulomb` core we need specific types for `UnitTypeName` and con `coulomb-test` as some tests are jvm only

coulomb/js/src/main/scala/coulomb/typename.scala
erikerlandson: As I understand it, any particular macro is one or the other.  So I'm guessing only one of whitebox or blackbox needs to be imported.

cquiroz: Ok, I'll clean this up a bit, I also thought I could move the `=:=` operation to another typeclass that is only used in testing

cquiroz: I don't need `blackbox` at all

erikerlandson:
regarding the `=:=` operator I'm using in the unit testing utilities, it may be sufficient to just test the `.name` field for string equality. We've already bound the `TypeName` implicits, so  like `typename1.name == typename2.name` ought to be equivalent to testing type equality.

cquiroz:
Right, I didn't want to change thecode too much, let me try comparing just name

cquiroz:
Yes, it works with just `.name`, I added another commit

coulomb-tests/jvm/src/test/scala/coulomb/SerializationTest.scala
```diff
@@ -0,0 +1,35 @@
...
+  val tests = Tests {
+    test("quantity be serializable") {
```
erikerlandson, line 19: should one of these be the Quantity serialize test

cquiroz: Do you refer to the class name or the name of the test? One of the comes form QuantityTest, the other form TemperatureTest

erikerlandson: they seem to contain the same code, both testing Temperature

cquiroz: Sorry my mistake while moving the tests, Fixed now

erikerlandson: It still looks the same, did you push the change?

erikerlandson: looking for quantity ser/de test:
```scala
    test("quantity be serializable") {
      import coulomb.scalatest.serde._
      val qs = Quantity[Int, Meter %/ Second](10)
      val qd = roundTripSerDe(qs)
      assert(qd.isValidQ[Int, Meter %/ Second](10))
      assert(qd === qs)
    }
```

cquiroz: Rebased onto develop and now it includes the correct test

erikerlandson:
LGTM, I'm going to merge. Thanks @cquiroz !

cquiroz:
Thanks, I learned quite a bit from this

PR #113, opened 2020-10-10
Add physical constants module

cquiroz:
This is a draft of adding a set of physical constants
There are several missing from the Wikipedia list but I wanted to get feedback before completing the rest

### DIFF
--- a/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
+++ b/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
@@ -69,10 +69,4 @@ package object mks {
   trait Siemens
   implicit val defineUnitSiemens = DerivedUnit[Siemens, (Second %^ 3) %* (Ampere %^ 2) %/ (Kilogram %* (Meter %^ 2))](abbv = "S")
 
-  trait ElectronVolt
-  implicit val defineUnitElectronVolt = {
-    val denom = Rational(10).pow(19)
-    val num = Rational(1602176634L) / Rational(10).pow(9)
-    DerivedUnit[ElectronVolt, Joule](abbv = "eV", coef = num / denom)
-  }
 }
--- a/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
+++ b/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
@@ -20,6 +20,7 @@ import singleton.ops._
 
 import coulomb.define._
 import coulomb.si._
+import spire.math.Rational
 
 /** MKS (Meter, Kilogram, Second) unit definitions (Newton, Joule, Watt, etc) */
 package object mks {
@@ -64,4 +65,11 @@ package object mks {
 
   trait Siemens
   implicit val defineUnitSiemens = DerivedUnit[Siemens, (Second %^ 3) %* (Ampere %^ 2) %/ (Kilogram %* (Meter %^ 2))](abbv = "S")
+
+  trait ElectronVolt
+  implicit val defineUnitElectronVolt = {
+    val denom = Rational(10).pow(19)
+    val num = Rational(1602176634L) / Rational(10).pow(9)
+    DerivedUnit[ElectronVolt, Joule](abbv = "eV", coef = num / denom)
+  }
 }
--- a/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
+++ b/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
@@ -27,6 +27,9 @@ package object mks {
   trait Radian
   implicit val defineUnitRadian = DerivedUnit[Radian, Unitless](abbv = "rad")
 
+  trait Steradian
+  implicit val defineUnitSteradian = DerivedUnit[Steradian, Unitless](abbv = "sr")
+
   trait Hertz
   implicit val defineUnitHertz = DerivedUnit[Hertz, Second %^ -1](abbv = "Hz")
 
--- a/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
+++ b/coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
@@ -61,4 +61,7 @@ package object mks {
 
   trait Henry
   implicit val defineUnitHenry = DerivedUnit[Henry, Weber %/ Ampere](abbv = "H")
+
+  trait Siemens
+  implicit val defineUnitSiemens = DerivedUnit[Siemens, (Second %^ 3) %* (Ampere %^ 2) %/ (Kilogram %* (Meter %^ 2))](abbv = "S")
 }
--- a/coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
+++ b/coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
@@ -62,23 +62,37 @@ package object physicalconstants {
 
   def neutronMass[V](implicit pcq: PhysicalConstantQuantity[V, NeutronMass]): Quantity[V, pcq.QU] = pcq.q
 
+  def bohrRadius[V](implicit pcq: PhysicalConstantQuantity[V, BohrRadius]): Quantity[V, pcq.QU] = pcq.q
+
+  def classicalElectronRadius[V](implicit pcq: PhysicalConstantQuantity[V, ClassicalElectronRadius]): Quantity[V, pcq.QU] = pcq.q
+
+  def electronGFactor[V](implicit pcq: PhysicalConstantQuantity[V, ElectronGFactor]): Quantity[V, pcq.QU] = pcq.q
+
+  def fermiCouplingConstant[V](implicit pcq: PhysicalConstantQuantity[V, FermiCouplingConstant]): Quantity[V, pcq.QU] = pcq.q
+
+  def hartreeEnergy[V](implicit pcq: PhysicalConstantQuantity[V, HartreeEnergy]): Quantity[V, pcq.QU] = pcq.q
+
+  def quantumOfReaction[V](implicit pcq: PhysicalConstantQuantity[V, QuantumOfCirculation]): Quantity[V, pcq.QU] = pcq.q
+
 }
 
 package physicalconstants {
+
+import coulomb.siprefix.Giga
 
   // actual unit defs in a subpackage
   object units {
     // define physical constants as derived units
     trait SpeedOfLightInVacuum
     implicit val defineSpeedOfLightInVacuum =
-      DerivedUnit[SpeedOfLightInVacuum, Meter %/ Second](coef = 299792458, name= "speed-of-light", abbv = "c")
+      DerivedUnit[SpeedOfLightInVacuum, Meter %/ Second](coef = 299792458, name = "speed-of-light", abbv = "c")
 
     trait PlanckConstant
     implicit val definePlanckConstant = {
       // original value = 6.62607015×10−34 J⋅s
       val denom = Rational(10).pow(34)
       val num = Rational(62607015) / Rational(10).pow(7)
-      DerivedUnit[PlanckConstant, Joule %* Second](coef = num / denom, name= "planck-constant", abbv = "ℎ")
+      DerivedUnit[PlanckConstant, Joule %* Second](coef = num / denom, name = "planck-constant", abbv = "ℎ")
     }
 
     trait ReducedPlanckConstant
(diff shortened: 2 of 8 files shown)


erikerlandson:
@cquiroz  this is cool, thanks!  It looks good to me but I'll review

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
```diff
@@ -0,0 +1,191 @@
...
+      val num = Rational(62607015) / Rational(10).pow(7)
+      DerivedUnit[PlanckConstant, Joule %* Second](coef = num / denom, name= "planck-constant", abbv = "ℎ")
+    }
+
+    trait ReducedPlanckConstant
+    implicit def defineReducedPlanckConstant = {
+      // original value = 1.054571817×10−34 J⋅s
+      val denom = Rational(10).pow(34)
+      val num = Rational(1054571817) / Rational(10).pow(9)
+      DerivedUnit[ReducedPlanckConstant, Joule %* Second](coef = num / denom, name= "reduced-planck-constant", abbv = "ℏ")
```
erikerlandson: I like the use of unicode abbreviations in this application :+1:

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
```diff
@@ -0,0 +1,191 @@
...
+
+}
+
+package physicalconstants {
+
+  // actual unit defs in a subpackage
+  object units {
+    // define physical constants as derived units
+    trait SpeedOfLightInVacuum
+    implicit def defineSpeedOfLightInVacuum =
```
erikerlandson: is there a reason you are creating the `DerivedUnit` instances using `def` (dynamic) instead of `val` or `object` (static)

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
```diff
@@ -0,0 +1,191 @@
...
+  package infra {
+
+    trait PhysicalConstantQuantity[V, CU] {
+      type QU
+      val q: Quantity[V, QU]
+    }
+    object PhysicalConstantQuantity {
+      type Aux[T, V, CU] = PhysicalConstantQuantity[V, CU] {type QU = T}
+
+      implicit def summon[V, CU, U](implicit
```
erikerlandson: I was starting to like using `summon` but it's going to be a keyword in scala 3 so probably should avoid it

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
erikerlandson: I guess it helps to use `denom` and `num`

cquiroz: I could change the to use val, I guess it will be slightly more efficient
I confronted how to create big numbers and didn't want to use doubles to avoid precision loss hence I turned to represent numbers using `Rational` operations

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
cquiroz: Good call

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
erikerlandson: `Rational` is how I prefer to do it.  Defer any precision loss to the "end" when the final numeric type is populated.

cquiroz:
All constants are now in place and I reviewed them so hopefully there will be no transcription errors.

Should be ready for a final review

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
```diff
@@ -0,0 +1,495 @@
...
+    implicit val defineClassicalElectronRadius = {
+      // original value = 2.8179403262(13)×10−15 m
+      val denom = Rational(10).pow(15)
+      val num = Rational(28179403262L) / Rational(10).pow(10)
+      DerivedUnit[ClassicalElectronRadius, Meter](coef = num / denom, name = "classical-electron-radius", abbv = "rₑ")
+    }
+
+    trait ElectronGFactor
+    implicit val defineElectronGFactor = {
+      // original value = −2.00231930436256(35)
```
cquiroz, line 321: This is a problem because you cannot pass a negative as coefficient.
I don't know if there is a way around it or whether I should just remove this constant

coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
```diff
@@ -61,4 +65,14 @@ package object mks {
 
   trait Henry
   implicit val defineUnitHenry = DerivedUnit[Henry, Weber %/ Ampere](abbv = "H")
+
+  trait Siemens
+  implicit val defineUnitSiemens = DerivedUnit[Siemens, (Second %^ 3) %* (Ampere %^ 2) %/ (Kilogram %* (Meter %^ 2))](abbv = "S")
+
+  trait ElectronVolt
```
cquiroz: This is worth discussing a bit. There are several units called electron volt referring to the same
thing but in terms of energy (this particular version) or length etc:
https://en.wikipedia.org/wiki/Electronvolt

erikerlandson: IMO the energy-based one you defined is the "right" one to have as the default.  If we ever added other variations, I'd qualify them. ElectronVoltMass, ElectronVoltDistance, etc.

Out of curiosity, why did you add ElectronVolt to 'mks' instead of physical constants?

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
erikerlandson: can you rename that to `summonConstant` or some such?

coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
cquiroz: I wasn't sure where to put it really. I think Wikipedia indicates it is an SI unit. I could move it to wherever you think it is the right place

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
```diff
@@ -0,0 +1,495 @@
...
+
+  def protonMass[V](implicit pcq: PhysicalConstantQuantity[V, ProtonMass]): Quantity[V, pcq.QU] = pcq.q
+
+  def neutronMass[V](implicit pcq: PhysicalConstantQuantity[V, NeutronMass]): Quantity[V, pcq.QU] = pcq.q
+
+  def bohrRadius[V](implicit pcq: PhysicalConstantQuantity[V, BohrRadius]): Quantity[V, pcq.QU] = pcq.q
+
+  def classicalElectronRadius[V](implicit pcq: PhysicalConstantQuantity[V, ClassicalElectronRadius]): Quantity[V, pcq.QU] = pcq.q
+
+  def electronGFactor[V](implicit pcq: PhysicalConstantQuantity[V, ElectronGFactor]): Quantity[V, pcq.QU] = pcq.q
```
erikerlandson: could add the negation here, just add a comment about it

erikerlandson: you will probably also have to add implicits to specify that `V` is AdditiveGroup.  I think that will qualify it as negatable.  If not that, there will be a way.

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
erikerlandson: We should be able to negate it up above in definition of `electronGFactor`

coulomb-mks-units/src/main/scala/coulomb/mks/package.scala
erikerlandson: good question.  It's not really an MKS unit, but it's not a physical _constant_ per se either.  I think I'd put it in `physicalconstants.units`

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
cquiroz: Renamed

coulomb-physical-constants/src/main/scala/coulomb/physicalconstants/package.scala
cquiroz: It seems I can make it work with `UnitNeg` as:
`def electronGFactor[V: UnitNeg](implicit pcq: PhysicalConstantQuantity[V, ElectronGFactor]): Quantity[V, pcq.QU] = -pcq.q`

erikerlandson: aha yes, even better

erikerlandson:
This LGTM. If I don't think of anything else I will merge later today or tomorrow

erikerlandson:
@cquiroz  I should have remembered this earlier - can you add a `README.md` with a couple examples of obtaining constants with differing value types to the new package?

erikerlandson:
LOL check this out, in scalaJS the engineering notation prints with "e" instead of "E":

```sh
$ sbt coulomb_physical_constantsJS/test

-------------------------------- Running Tests --------------------------------
t= 6.6743e-11 m^3/(kg s^2)
X coulomb.physical.PhysicalConstantsSuite.physical constants 40ms 
  utest.AssertionError: assert(newtonianConstantOfGravitation[Double].show == "6.6743E-11 m^3/(kg s^2)")
```

This mostly seems to be happening for Float and Double (not BigDecimal).
I didn't see this but Travis wasn't running new tests because they aren't defined in `coulomb_tests`. It looks like you can fix it with:

```scala
  implicit class EnhanceTShow[V, U: UnitString](q: Quantity[V, U]) {
    def tshow: String = {
      val vstr = q.value.toString.toUpperCase
      val ustr = q.showUnit
      s"$vstr $ustr"
    }
  }

  // ...

assert(newtonianConstantOfGravitation[Double].tshow == "6.6743E-11 m^3/(kg s^2)")
```

If you can move the physical constant tests under `coulomb_tests` so travis will run them and relace `.show` with `.tshow` it should get all the tests running and it will be mergeable

cquiroz:
https://github.com/scala-js/scala-js/issues/3202

cquiroz:
It is very difficult to get a `tshow` implementation that will work in all cases

I could put a parallel js test or in each test accept either format like

`assert(newtonianConstantOfGravitation[Double].show == jvmformat || newtonianConstantOfGravitation[Double].show == jsformat)`

erikerlandson:
I see, it's worse than just `e` vs `E`, it's not even consistent with using scientific vs straight decimals.  I think testing based on the string representations is going to be problematic.

Here's an alternative approach where you can directly test the types and values.

```scala
object PhysicalConstantsSuite extends TestSuite {

  implicit class WithEqQMEqOp[V, U](q: Quantity[V, U]) {
    // requires exact V & U types of 'q', which removes the ambiguities of "smart" conversions
    // and makes sure we are getting exactly the type we defined.
    def =?=(qt: Quantity[V, U]): Boolean = {
      val r = (q.value == qt.value)
      if (!r) {
        // help debug test when it fails
        println(s"${q.value} != ${qt.value}")
      }
      r
    }
  }

  val tests =  Tests {
    test("physical constants") {
      assert(speedOfLightInVacuum[Int] =?= Quantity[Int, Meter %/ Second](299792458))
    }
  }
}
```

erikerlandson:
One last variation, where `.tshow` is defined using fprintf style to control the output.

```scala
  implicit class EnhanceTShow[V, U: UnitString](q: Quantity[V, U])(implicit cf: ConvertableFrom[V]) {
    def tshow: String = {
      val vf = cf.toType[Double](q.value)
      val vstr = f"${vf}%.8E"
      val ustr = q.showUnit
      s"$vstr $ustr"
    }
  }

  val tests =  Tests {
    test("physical constants") {
      //assert(speedOfLightInVacuum[Int] =?= Quantity[Int, Meter %/ Second](299792458))
      val t = speedOfLightInVacuum[Int].tshow
      println(t)
      assert(t == "2.99792458E+08 m/s")
```

cquiroz:
I agree about testing with String. One of the reasons to do so was to let me easily compare with the values in wikipedia.
Also using `show` revealed a problem with the dependent types in `sumonConstant`

cquiroz:
I moved the tests to be executed by travis and now use a test without `show`. I preserved the test with String for the jvm side as it maybe useful but I could remove it if you prefer

erikerlandson:
LGTM, merging.  Thanks @cquiroz !

PR #124, opened 2020-11-12
Port to munit

### LINKED ISSUE #123: Switch to munit
What do you think about switching to `munit` for tests. The main benefit is that we can get dotty support sooner.

If you're ok with that I could prepare a PR

cquiroz:
Fixes #123


erikerlandson:
maybe was off base on "smaller deps"

cquiroz:
What do you mean? Because the size change? The switch is fairly mechanical but the diff is indeed large

erikerlandson:
Just that munit is using 3 dependencies instead of 1.  Which isn't objectively bad.  I thought that being right on top of JUnit would be "smaller".

cquiroz:
Right there is an adapter for scalacheck tests, I think they are only needed for the cats module

erikerlandson:
travis is picking up an import error on `import utest._`
I need to get CI ported to github actions or something

cquiroz:
Travis is kind of removing support for OSS projects, I have ported a bunch to GHA and could do the same for coulomb

erikerlandson:
If you want to I won't say no :+1:

erikerlandson:
looks like the gh build ci is working, thank you! - can you rebase this to pick up the CI

cquiroz:
Done

cquiroz:
Uups, I missed one test

Issue #147, opened 2020-12-24
Remove dependency on spire


erikerlandson:
The approach would be to port spire's `Rational` into the `coulomb` tree. The typeclasses for `AdditiveGroup` and friends can be obtained from `typelevel/algebra`.  The `spire` package could then become `%Test`

erikerlandson:
The goal is to avoid direct dependency on spire so that it does not block port to Scala 3.  Having it be `%Test` is not a total avoidance, but it would allow continued testing in Scala 2, and testing in Scala 3 could just use native numeric types to start, plus in-tree Rational

erikerlandson:
One option would be to implement a sort of MVP rational that is not necessarily a direct port of spire Rational. As used by `coulomb` core, the Rational is compile-time only in the implicit system, so less-than-maximum efficiency is lower impact.

cquiroz:
I guess that is an option. I'm trying to get `spire` to run in dotty but I don't have an ETA

erikerlandson:
The three main things coulomb currently consumes from spire are:
1. `Rational`
2. `ConvertableTo` and `ConvertableFrom`
3. `MultiplicativeGroup` and friends

`Rational` would be easy to replace internally.  `MultiplicativeGroup` and friends are being passed thru from typelevel `algebra`.  I think the trickiest one to replace would be `ConvertableTo` and `ConvertableFrom`.

It would be feasible to define a coulomb typeclass for `ConvertableTo` and `ConvertableFrom`, and simply pass-thru the spire version whereever that was in scope.

My two main concerns with `spire` are that (1) it took a long time for spire to get onto scala-2.13, and I hope to not have it block coulomb for scala 3, and (2) spire defines an awful lot of stuff, and I'm only using a couple things.

Regardless of whether it is a hard dependency, I want coulomb to always integrate seamlessly with spire numeric types, for obvious reasons.

erikerlandson:
side note: I'm also using `shapeless` for `::` and `HNil`, which I'm getting transitively from `spire` but these types are purely implementation details, and they would be trivial to replace with my own version of `::` and `HNil` internally

erikerlandson:
cross-ref: over on my scala-3-POC I have an in-tree Rational and fractional exponents:

https://github.com/erikerlandson/poc-coulomb-scala3/blob/cbf290521792974b0f78767495a8caa09fdcc974/coulomb/src/coulomb/infra/meta.scala#L231

erikerlandson:
[...]

Issue #130, opened 2020-11-17
scala 3 roadmap


erikerlandson:
This will be my tracking issue for figuring out how to approach scala 3

### branching
coulomb for scala 3 will start as a PR, but once it merges I'm going to make it the main (develop) branch going forward, and whatever the most recent version is with scala 2, that will become a maintenance-only branch.

### scala 2 compatibility mode
I may try building with scala-2/3 compatibility mode as an interim - that is, making no code changes but compiling for scala 3 as part of the crossbuild.  If crossbuild works well enough, this phase can be part of existing release series, with no maintenance branching needed.

### implicit -> summon
The main syntactic changes will be around contextual polymorphism, for example the new `summon` keyword. I expect this to be a fairly mechanical change to the existing code but it will be very pervasive. Once this change takes place, previous release will be its own maintenance branch.

### QuantityParser
I'll need to figure out if the existing implementation of QuantityParser will survive the transition to scala 3. See also #99.

### Scala 3 metaprogramming
Scala 2 macro paradise was hard to work with and somewhat deprecated. Scala 3 metaprogramming appears to be a big improvement and is a first-class language feature. One thing to do is look for opportunities to re-incorporate metaprogramming if that makes sense.

### other scala 3 features?
If there are any other scala 3 language features coulomb can leverage, they can be incorporated after the large-scale `implicit` -> `summon` changes.

cquiroz:
I have been experimenting a bit with this. One of the stumbling blocks maybe getting `spire` for dotty.

erikerlandson:
@cquiroz agreed, spire was also a blocker with scala 2.13 for a while. At one point I helped with the 2.13 build for spire because I couldn't proceed without it.

erikerlandson:
Scala 3 language features are substantially different in the space of context parameters ("implicits") and new type features, and so for my own clarity and sanity I am going to start experimenting and prototyping what the core of coulomb could look like when taking full advantage of Scala 3 in a separate repo:

https://github.com/erikerlandson/poc-coulomb-scala3

When I feel comfortable with how Scala 3 works, I can start a real scala-3 dev branch for coulomb.

erikerlandson:
xref: https://github.com/lampepfl/dotty/issues/11827

erikerlandson:
Quantity as an opaque type:

```scala
trait Addable[V]:
    def plus(x: V, y: V): V
  
given Addable[Int] with
    def plus(x: Int, y: Int): Int = x + y

[...]

PR #230, opened 2022-01-28
Update spire to 0.18.0-M3

armanbilge:
It's not a stable release yet, but this seems reasonable for the develop branch. This is a small first step towards https://github.com/erikerlandson/coulomb/issues/130, although I am aware there are alternate ideas e.g. https://github.com/erikerlandson/coulomb/issues/147.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -55,7 +55,7 @@ def commonSettings = Seq(
     Resolver.sonatypeRepo("snapshots")
   ),
   libraryDependencies ++= Seq(
-    "org.typelevel" %%% "spire" % "0.17.0" % Provided,
+    "org.typelevel" %%% "spire" % "0.18.0-M3" % Provided,
     "eu.timepit" %%% "singleton-ops" % "0.5.2" % Provided,
     "org.scalameta" %%% "munit" % "0.7.29" % Test,
     "org.typelevel" %%% "discipline-munit" % "1.0.9" % Test,


erikerlandson:
@armanbilge thanks!

armanbilge:
@erikerlandson thanks for merging! I'm going to be working on the Scala 3 port of coloumb for the folks at @gemini-hlsw. I understand you are essentially imagining a ground-up rewrite?

erikerlandson:
@armanbilge   yes, in fact this weekend I moved my "poc" code onto an official development branch `scala3`:
https://github.com/erikerlandson/coulomb/tree/scala3
cc @cquiroz 

This is an "orphan" branch - it cannot merge as a pull request against the original `develop` branch, but it has nothing in common with it, so I think that reflects reality.  My plan is to eventually make `scala3` the new `develop` and the original `develop` will become a scala-2.13 maintenance branch.

My thought is that if people want to contribute, then we can submit pull-reqs against `scala3`. 

The `mill` build.sc file is just a skeleton right now but I prototyped examples of everything I think we need to do a proper build structure here:
https://github.com/erikerlandson/mill-testbed/blob/main/build.sc

It demonstrates
- jvm+js cross builds
- basic subproject dependencies
- sonatype publishing
- "sbt-unidoc-style" microsite publishing
- unit testing (even heterogeneous test platforms)

armanbilge:
@erikerlandson thanks for the response. Exciting!

>My plan is to eventually make scala3 the new develop and the original develop will become a scala-2.13 maintenance branch.

👍 this sounds good to me.

I suppose there is little chance of preserving some semblance of source-compatibility across 2.13/3? This seems difficult although [scodec](https://github.com/scodec/scodec) has managed this with essentially the same setup. Without source-compatibility this means every downstream will also have to split their sources or create a new branch for Scala 3, but that may be unavoidable.

Re mill: of course, it is completely up to you how you run your build :)

IMHO I would stick to sbt, unless you have a specific reason/need to use mill. For better or for worse (certainly for worse, really), mill is still the minority build tool in the OSS space and thus it adds an overhead for contributors who are not familiar with it, as well as maintainers since the plugin ecosystem is less-developed.

Along those lines, I recently created the [sbt-typelevel](https://typelevel.org/sbt-typelevel/) plugin. It's currently being used by [24 typelevel libs and counting](https://github.com/typelevel/sbt-typelevel/issues/83), the http4s org, and I'm in the process of migrating the @gemini-hlsw builds as well.

Out-of-the-box it supports the following, with very minimal configuration:
- early semantic versioning
- mima binary-compatibility checking
- sonatype publishing in CI
- jvm+js cross builds
- microsites
- automatically-generated GitHub actions workflows

Versioning and binary-compatibility in particular can be non-trivial to get right and are extremely important for libraries. If you're open to considering it, I'd be happy to make a PR demonstrating what your build would look like using sbt-typelevel.

Thanks again, I'm looking forward to helping out with this effort! :)

erikerlandson:
Let's move this interesting discussion to #233

PR #234, opened 2022-02-07
Setup sbt-typelevel build

armanbilge:
Let's see how this goes :)

Note that the GitHub Actions workflows are completely auto-generated. All the actual config is in `build.sbt`.

I also added a Native cross-build, because, why not? :) I can remove it, or we can give it a try so long as it's not a problem.

Let me know if you have any questions!


build.sbt
```diff
@@ -0,0 +1,34 @@
+ThisBuild / tlBaseVersion := "0.6"
+
+// publish settings
+ThisBuild / developers += tlGitHubDev("erikerlandson", "Erik Erlandson")
+ThisBuild / organization := "com.manyangled"
+ThisBuild / organizationName := "Erik Erlandson"
+ThisBuild / startYear := Some(2022)
+
+// ci settings
+ThisBuild / tlCiReleaseBranches := Seq("scala3")
```
armanbilge, line 10: This enables CI-released snapshots from the `scala3` branch.

PR #236, opened 2022-02-09
Update build.sbt

armanbilge:


build.sbt
```diff
@@ -4,7 +4,6 @@ ThisBuild / tlBaseVersion := "0.6"
 ThisBuild / developers += tlGitHubDev("erikerlandson", "Erik Erlandson")
 ThisBuild / organization := "com.manyangled"
 ThisBuild / organizationName := "Erik Erlandson"
-ThisBuild / licenses := List("Apache-2.0" -> url("http://opensource.org/licenses/Apache-2.0"))
```
armanbilge, line 7: Keeping this is harmless of course, but I omitted it because it's already the default :)

armanbilge: Oh, unless you prefer that specific URL. sbt-typelevel uses `https://www.apache.org/licenses/LICENSE-2.0.txt`

erikerlandson: I'd like to keep the license setting, however `https://www.apache.org/licenses/LICENSE-2.0.txt` seems like a better url since it is from apache.org

build.sbt
```diff
@@ -36,7 +35,7 @@ lazy val all = project
   .in(file("all")) // sbt will create this - it is unused
   .dependsOn(core.jvm, units.jvm) // scala repl only needs JVMPlatform subproj builds
   .settings(name := "coulomb-all")
-  .settings(publish := {}, publishLocal := {}) // don't publish
+  .enablePlugins(NoPublishPlugin) // don't publish
```
erikerlandson, line 39: nice :+1:

erikerlandson (changes requested):

build.sbt
armanbilge: Done!

erikerlandson (approved):

PR #245, opened 2022-02-20
move unit declaration parameters into types

erikerlandson:
This significantly simplifies and optimizes the `Expr` objects generated for `Coefficient` and `Show` in the metaprogramming, and also makes some of the metaprogramming logic cleaner.

Unit defs now look like:

```scala
final type Meter
given ctx_Unit_Meter: BaseUnit[Meter, "meter", "m"] with {}

final type Kilo
given ctx_Unit_Kilo: PrefixUnit[Kilo, 10 ^ 3, "kilo", "k"] with {}

final type Minute
given ctx_Unit_Minute: DerivedUnit[Minute, Second, 60, "minute", "min"] with {}
```

### DIFF
--- a/core/src/main/scala/coulomb/define/define.scala
+++ b/core/src/main/scala/coulomb/define/define.scala
@@ -16,10 +16,34 @@
 
 package coulomb.define
 
+/**
+ * @tparam Name unit name
+ * @tparam Abbv unit abbreviation
+ */
 abstract class NamedUnit[Name, Abbv]
 
+/**
+ * @tparam U unit type
+ * @tparam Name unit name
+ * @tparam Abbv unit abbreviation
+ */
 abstract class BaseUnit[U, Name, Abbv] extends NamedUnit[Name, Abbv]
 
+/**
+ * @tparam U unit type
+ * @tparam D unit it is derived from
+ * @tparam Coef unit coefficient, relative to `D`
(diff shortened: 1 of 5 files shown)


core/src/main/scala/coulomb/define/define.scala
```diff
@@ -16,29 +16,10 @@
...
-abstract class DerivedUnit[U, D] extends NamedUnit:
-    import coulomb.infra.*
-    val coef: Rational
-    override def toString = s"DerivedUnit($coef, $name, $abbv)"
-
-// Not necessary, but allows meta-programming to be smarter with coefficients
-abstract class DerivedUnit1[U, D] extends DerivedUnit[U, D]:
-    val coef: Rational = Rational.const1
-
-// prefix units are derived units of 1 ('unitless')
```
armanbilge, line 43: We lost some of these helpful comments :)

armanbilge (approved):
The metaprogramming changes are over my head :) but the new definitions of `BaseUnit` and friends are extremely nice!

core/src/main/scala/coulomb/define/define.scala
armanbilge: - https://github.com/erikerlandson/coulomb/pull/246

PR #251, opened 2022-03-04
Publish a unidoc artifact

armanbilge:
This sets up a new published artifact `coulomb-docs` that contains the aggregated unidocs for the project.

javadoc.io doesn't render SNAPSHOT releases, so it won't do much until a real release. However I did preview locally and they look good!


erikerlandson:
Thanks, merging!

Issue #250, opened 2022-03-03
Scala 3 tests are not running on JS or Native


armanbilge:
This is because of the `%%` instead of `%%%`.

https://github.com/erikerlandson/coulomb/blob/5c9aa623ddfbc0b2e9d9ed405dd37983ffc9846e/build.sbt#L27

Actually, at the moment it's impossible to run the tests at all on Native, since the munit support is blocked:
- https://github.com/scalameta/munit/pull/477

We have a couple options:
1. Keep the status quo for now, since the `%%` effectively disables the platform tests anyway (maybe this was intentional? :)
2. Temporarily drop Native, so we can at least enable tests on JS which is an important platform for the folks at @gemini-hlsw.

erikerlandson:
It's semi-deliberate - I tried `%%%` and as you say ran into problem with native.  I wasn't sure what the cleanest way to proceed was, so I put my head in the sand.

I definitely want to support JS for @cquiroz, so if native is blocked it's reasonable to just not build native for now, although it would be cool to do so in the future.

armanbilge:
Yes, I think it's good to start running tests on JS as soon as possible. I think the JVM and Native platforms are more similar, whereas JS has its own quirks, such as the often surprising all-numbers-are-actually-`Double` semantics :)

I anticipate the native stuff will become unblocked even within a matter of weeks :)

PR #252, opened 2022-03-04
Enable JS tests, drop Native for now

armanbilge:
Closes https://github.com/erikerlandson/coulomb/issues/250.


armanbilge:
Btw, in theory we could hack the build to work around this, but it's definitely not worth it 😁

PR #253, opened 2022-03-04
Add a testkit

armanbilge:
Anything else that I should add?


core/src/main/scala/coulomb/rational/rational.scala
```diff
@@ -41,7 +41,6 @@ final class Rational private (val n: BigInt, val d: BigInt) extends Serializable
         if (e < 0) then
             canonical(d.pow(-e), n.pow(-e))
         else if (e == 0) then
-            require(n != 0)
             canonical(1, 1)
         else if (e == 1) then
             this
```
armanbilge, line 46: According to algebra, this was a bug.

armanbilge:
Sorry, I ended up adding a testkit with ScalaCheck instances as well :)

core/src/main/scala/coulomb/rational/rational.scala
```diff
@@ -78,6 +78,8 @@ final class Rational private (val n: BigInt, val d: BigInt) extends Serializable
         case v: Long => (n == v) && (d == 1)
         case _ => false
 
+    override def hashCode: Int = 29 * (37 * n.## + d.##)
```
armanbilge, line 81: Another bug, caught by laws. I copied this from spire.

https://github.com/typelevel/spire/blob/b8d5c71b499e7e477a0abed1db344298b31c3be5/core/src/main/scala/spire/math/Rational.scala#L737-L739

erikerlandson:
Thanks, I just saw this!  I'll take a deeper look when I have a bit of time.

A couple things - 
distinction between MultiplicativeGroup and Truncating Division
cases for `.pow` across multiplicative semigroup, field, "support for roots" (still looking at this myself)

armanbilge:
Not sure if this answers your questions, but:
- `TruncatedDivision` is the modulus operator, e.g. `%` not `/`
- Spire has an [`NRoot`](https://github.com/typelevel/spire/blob/b8d5c71b499e7e477a0abed1db344298b31c3be5/core/src/main/scala/spire/algebra/NRoot.scala#L30) typeclass
- not sure what you meant by this one :)
  > cases for .pow across multiplicative semigroup, field

erikerlandson:
`CanAdd` and friends are currently an implementation detail of `coulomb.ops.standard`, and one I don't actually think I need except possibly for `Pow` - you can see that `CanAdd`, `CanSub`, `CanMul` and `CanNeg` are all just direct calls to `Numeric`, and I should probably just remove them and refer directly to `Numeric` in the various rules.  `CanDiv` has only two cases, for `Fractional` and `Integral`, which will also carry over into `algebra` in the form of `MultiplicativeGroup` and `TruncatingDivision`

My vision was to "copy" the rules under `coulomb.ops.standard`, but replace `using alg: ...` with various typeclasses from `algebra` package.  I was not planning to carry over `CanXXX`.  I was also considering not carrying over the specialized `Double` and `Float` rules, because those technically represent an opinion about the underlying algebra.  I might put these optimizations in a separately-importable package under spire, haven't decided.

`CanPow` and power in general is an interesting case.  There is no `scala.math.Numeric` support for power, so I wrote my own in `coulomb.ops.standard`.  In `algebra` they allow `.pow(<positive-int>)` for `MultiplicativeSemigroup`, and `.pow(<int>)` for `MultiplicativeGroup`.  I also intend to support `.pow(<rational>)`, and I have not seen any algebra typeclass that supports this: support for something like `.pow(<double>)` OR `.pow(<int>) and .root(<int>)` would be sufficient to do it.

armanbilge:
Hmm, I'm not entirely sure if I follow what you have in mind :(

> My vision was to "copy" the rules

Maybe I misunderstand what "copy" means, but this doesn't sound like a very DRY strategy.

IMHO, I would ideally embrace the algebra dependency in coulomb core. FTR, what is the argument against that? It already provides solid typeclasses backed by mathematical theory. Furthermore, it provides laws and discipline tests.

Regarding powers and roots, I really think what you are looking for is [`NRoot`](https://github.com/typelevel/spire/blob/b8d5c71b499e7e477a0abed1db344298b31c3be5/core/src/main/scala/spire/algebra/NRoot.scala#L30). It currently lives in spire, but it seems reasonable that we should port it to algebra (I'm a maintainer for both projects).

erikerlandson:
>  this doesn't sound like a very DRY strategy.

In some respects, it certainly isn't.  I have absolutely gone back and forth in my head about just directly using `algebra` in `coulomb-core`, probably 50 times in the last week :joy: 

It's a relatively light weight dep, and I definitely don't hate the idea.  I think the reasons I have hesitated are: 
1. My thinking around the interaction of `Div`, `Pow` and `AllowTruncation` wasn't solidified yet.  It's more solidified now.   
2.  I had mixed feeling about the `double` specific rules and generalized `algebra`, but I DO wish to allow support for type-specific optimizations like that.  
3.  The somewhat inconsistent levels of support for `BigDecimal` and `BigInt` - like, they appear to be supported in some algebras, but NOT in scala's native conversions - they have their own special `ScalaNumericConversion` typeclass and there is no typeclass support for things like `double -> BigDecimal` that I can find. 
4.   there is still a part of me that loves the idea of true zero deps for `core`

Maybe the answer is:
1. suck it up and consume `algebra` directly in `coulomb-core`
2. specialized "optimized" rules for `Double` and `Float` (maybe others eventually?) go in their own importable `coulomb.ops.standard.optimizations` or something.   `coulomb-spire` will probably add to these
3. Embrace the inconsistencies around `scala.math` and `BigDecimal`, etc - they are what they are, and one will be able to get better, more consistent policy options from `coulomb-spire`
4. a single light-weight dep that adds significant value is smart design compromise, and it's part of the typelevel ecosystem anyway.
5.  Yes, please move `NRoot` to `algebra` - if they can have `TruncatingDivision` they can certainly have `NRoot` :grin: 

So in this concept, there will be no `coulomb-algebra`, that will be `coulomb-core`.  `coulomb-spire` will probably be mostly alternative (and better) conversion rules using their `ConvertibleFrom` typeclass, plus some other implementation support for unit conversions, value-resolution, etc.

I'll create a pullrequest and show it to you and @cquiroz

erikerlandson:
now that we moved `algebra` into coulomb-core I think this is no longer necessary

armanbilge:
This PR also added algebra instances for rational, scalacheck arbitraries, and discipline-based tests which helped catch some bugs in rational :)

erikerlandson:
re-opened :+1:

build.sbt
```diff
@@ -43,12 +43,39 @@ lazy val units = crossProject(JVMPlatform, JSPlatform/*, NativePlatform*/)
...
+  .in(file("testkit"))
+  .settings(
+    name := "coulomb-testkit",
+    libraryDependencies ++= Seq(
+      "org.scalacheck" %%% "scalacheck" % "1.15.4",
+      "org.scalameta" %%% "munit-scalacheck" % "0.7.29" % Test,
+    )
+  )
+  .dependsOn(core % "compile->compile;test->test")
+  .settings(commonSettings :_*)
```
armanbilge: @erikerlandson would you be open to accepting this `testkit` module? Even if you don't test laws internally we'd find it helpful in our projects.

erikerlandson: sure, law checking is good CI

core/src/main/scala/coulomb/rational/rational.scala
armanbilge: @erikerlandson how do you feel about these changes to `Rational`? I noticed you didn't cherry-pick them.

erikerlandson: Yes, that is a bug!

build.sbt
```diff
@@ -54,6 +54,19 @@ lazy val spire = crossProject(JVMPlatform, JSPlatform, NativePlatform)
   .settings(commonSettings :_*)
   .settings(libraryDependencies += "org.typelevel" %%% "spire" % "0.18.0")
 
+lazy val testkit = crossProject(JVMPlatform, JSPlatform, NativePlatform)
```
erikerlandson, line 57: @armanbilge Should the `testkit` subproject have a `NoPublishPlugin` ?

armanbilge: The idea is to publish it, so we can use it downstream in our projects.

core/src/main/scala/coulomb/rational/rational.scala
erikerlandson: :+1:

build.sbt
erikerlandson: ah, I see.

erikerlandson:
@armanbilge  it LGTM - if you are good with it, I'll merge

armanbilge:
Since you're happy with the changes to `Rational`, let me add back the law tests that let me catch them originally. Then we should be good to go. Thanks for the review!

armanbilge:
@erikerlandson should be good to go! :)

PR #255, opened 2022-03-13
direct use of scala.Cconversion in operations

erikerlandson:


core/src/main/scala/coulomb/conversion/standard/implicitConversion.scala
```diff
@@ -27,7 +27,29 @@ object implicitConversion:
     // I gate this with AllowImplicitConversions because many operations on coulomb
```
cquiroz: is `AllowImplicitConversions` part of coulomb or scalalib?

erikerlandson: `AllowImplicitConversions` it is a coulomb thing I defined.

Scala has `scala.language.implicitConversions` which you can import, but it isn't an implicit var so I can't use it.

erikerlandson:
both the current system and this experiment have pros and cons.

current:
- pro: cleaner optimization story
- con: use of `AllowImplicitConversions` is a bit awkward

experiment:
- pro: works with the new `scala.Conversion` convention more correctly
- con: optimization not quite as good, uses redundant throw-away `Conversion` in optimized rules.

erikerlandson:
Benchmarking experiment
I ran the following four `foldLeft` additions for:
- `+` (current),
- `++` (experiment)
- `+` (current optimized)
- `++` (experiment optimized)
- `++` (using explicit conversions `toValue` and `toUnit`

The data vector to sum was generated like so:
```scala
scala> val data = Vector.fill(10000000) { math.random().withUnit[Meter] }
val data: Vector[coulomb.quantity.Quantity[Double, coulomb.units.mks.Meter]] = Vector(0.17273950018557938 ...
```
For each of the five scenarios listed above, I ran the following four variations:
- 1V1U "value and unit type unchanged"
- 1V2U "value unchanged, unit changed"
- 2V1U "value changed, unit unchanged"
- 2V2U "value and unit type both changed"

An example of each of these for, for `+`:
```scala
scala> benchmark.medianTime({ data.foldLeft(0d.withUnit[Meter]){ (z, x) => z + x } }, n=51)
val res36: (Long, coulomb.quantity.Quantity[Double, coulomb.units.mks.Meter]) = (159,4999411.42929234)
                                                                                                                                                               
scala> benchmark.medianTime({ data.foldLeft(0d.withUnit[1000 * Meter]){ (z, x) => z + x } }, n=51)
val res37: (Long, 
  coulomb.quantity.Quantity[Double, 1000 * coulomb.units.mks.Meter]
) = (161,4999.41142929375)
                                                                                                                                                               
scala> benchmark.medianTime({ data.foldLeft(0f.withUnit[Meter]){ (z, x) => z + x } }, n=51)
val res38: (Long, coulomb.quantity.Quantity[Float, coulomb.units.mks.Meter]) = (143,4999385.5)
                                                                                                                                                               
scala> benchmark.medianTime({ data.foldLeft(0f.withUnit[1000 * Meter]){ (z, x) => z + x } }, n=51)
val res39: (Long, 
  coulomb.quantity.Quantity[Float, 1000 * coulomb.units.mks.Meter]
) = (141,4996.872)
```
Times above collected in milliseconds.  Each reported time is the median of 51 trials.

The resulting 5x4 matrix of times:
image

A plot of these values:
image

Discussion:
1. We can see from the data and plot that the current version of operators (using `ValueConversion` and `UnitConversion` directly, without `scala.Conversion`) is remarkably consistent whether the `Double` optimization rules are enabled or not (blue and yellow lines in the plot)  In fact, if I decide to keep the current system, I now suspect the optimization rules are unnecessary.
2. In the experimental implementation (using `scala.Conversion`), adding changes of unit and value types to convert has a very significant effect on time, without optimizations (red line).
3. The optimizing rules have a big impact on the experimental implementation (green line).  Note, only the first two categories are optimized (1V1U, 1V2U).  Optimizing the other cases is possible, but runs into combinatoric cases of value conversions, and my personal view is that the first two cases are the most useful to optimize anyway.
4. Using explicit conversion calls `.toValue` and `.toUnit` is also a very effective way to optimize performance (orange line).

Based on these data, I'm feeling more comfortable with moving forward with the experimental `scala.Conversion` based implementation: I can optimize the common cases I believe are most relevant, and users who wish to optimize other cases can either write their own customized implicit optimizations, or simply invoke `.toValue` and `.toUnit` directly in their code.  Future improvements to scala compiler, introduction of `erased` keyword, or improvements enabled by java Valhalla, can only improve performance options in coulomb.

Appendix
- an example of explicit use of `.toValue` or `.toUnit` (orange line in plot):
```scala
scala> benchmark.medianTime({ data.foldLeft(0f.withUnit[1000 * Meter]){ (z, x) => z ++ x.toValue[Float].toUnit[1000 * Meter] } }, n=51)
val res52: (Long, 
  coulomb.quantity.Quantity[Float, 1000 * coulomb.units.mks.Meter]
) = (141,4996.872)
 ```
- Optimizations for `Double` were enabled by
```scala
scala> import coulomb.ops.standard.optimizations.double.given
```

erikerlandson:
Addendum

The results in the experiment above were obtained importing *only* the Double specific optimizing rules:
```scala
import coulomb.ops.standard.optimizations.double.given
```

When I import the other rules: 
```scala
import coulomb.ops.standard.optimizations.all.given
```

the performance picture is even better.  For the "experiment-optimized" case the numbers are now:
- 1V1U: 126ms
- 1V2U: 137
- 2V1U: 143
- 2V2U: 155

So the experimental concept using `scala.Conversion` is fully competitive with the current version, at least with both Float and Double optimizations in scope

PR #261, opened 2022-03-23
Demonstrate issue with exported types

armanbilge:
Not for merging :)

This PR creates a test case with the compiler issue reported in https://github.com/erikerlandson/coulomb/discussions/237#discussioncomment-2411618.

Then, it fixes it by replacing exported types with type aliases.

So, either this is a compiler bug, or exporting types doesn't do what we'd expect 😅

### DIFF
--- a/core/src/main/scala/coulomb/deltaquantity.scala
+++ b/core/src/main/scala/coulomb/deltaquantity.scala
@@ -16,7 +16,8 @@
 
 package coulomb
 
-export deltaquantity.DeltaQuantity as DeltaQuantity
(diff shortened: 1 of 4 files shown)


erikerlandson:
Exporting types actually seems to be working very well.   The problem is highly localized to two methods: `show` and `showFull`, and furthermore the problem is only occurring when the type is being mediated thru a type def, for example via `type Temperature[V, U] = ...`.  Note that mediating via `Temperature` (or `EpochTime`, etc) is not causing problems for any other methods, only these two.

PR #264, opened 2022-03-31
experiment with removing `transparent inline` from op definitions

erikerlandson:
An experiment in removing `transparent inline` on operators.

This branch removes `transparent inline` from the `pow` method.

### DIFF
--- a/core/src/main/scala/coulomb/ops/standard/pow.scala
+++ b/core/src/main/scala/coulomb/ops/standard/pow.scala
@@ -26,18 +26,19 @@ import coulomb.ops.{Pow, SimplifiedUnit}
 import coulomb.rational.typeexpr
 import coulomb.ops.algebra.FractionalPower
 
-transparent inline given ctx_pow_FractionalPower[V, U, E](using
+given ctx_pow_FractionalPower[V, U, E](using
     alg: FractionalPower[V],
+    dbv: typeexpr.DoubleValue[E],
     su: SimplifiedUnit[U ^ E]
         ): Pow[V, U, E] =
-    val e = typeexpr.double[E]
+    val e = dbv.value
     new Pow[V, U, E]:
         type VO = V
         type UO = su.UO
         def apply(q: Quantity[V, U]): Quantity[VO, UO] =
             alg.pow(q.value, e).withUnit[UO]
 
-transparent inline given ctx_pow_MultiplicativeGroup[V, U, E](using
+given ctx_pow_MultiplicativeGroup[V, U, E](using
     nfp: NotGiven[FractionalPower[V]],
     alg: MultiplicativeGroup[V],
     aie: typeexpr.AllInt[E],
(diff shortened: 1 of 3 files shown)


erikerlandson:
The results are interesting.  The operator itself, in isolation, works, but it does lose important type information.

```scala
Welcome to Scala 3.1.1 (11.0.14, Java OpenJDK 64-Bit Server VM).
                                                                                                                                            
scala> import coulomb.*, algebra.instances.all.given, coulomb.ops.standard.given, coulomb.policy.standard.given
                                                                                                                                            
scala> import coulomb.units.us.{*,given}, coulomb.units.mks.{*,given}, coulomb.units.accepted.{*,given}
                                                                                                                                            
scala> 2d.withUnit[Meter].pow[3]
val res0: Any = 8.0
                                                                                                                                            
scala> res0.asInstanceOf[Quantity[Double, Meter ^ 3]]
val res1: coulomb.Quantity[Double, coulomb.units.us.Meter ^ 3] = 8.0
                                                                                                                                                                                                                                                                                        
scala> def f(q: Quantity[Double, Meter ^ 3]): Double = q.value
def f(q: coulomb.Quantity[Double, coulomb.units.us.Meter ^ 3]): Double
                                                                                                                                            
scala> f(res0)
-- [E007] Type Mismatch Error: -------------------------------------------------
1 |f(res0)
  |  ^^^^
  |  Found:    (res0 : Any)
  |  Required: coulomb.Quantity[Double, coulomb.units.us.Meter ^ (3 : Int)]

longer explanation available when compiling with `-explain`
1 error found
                                                                                                                                            
scala> 
```

From the above we can see that applying the `pow` operator succeeds, but without `transparent inline` the type of the operator is `Any` and so the `coulomb` unit type system isn't propagated in a useful way in operation outputs.

This is also reflected in the failures of the unit tests.

erikerlandson:
Overall this does not seem viable.  It might be possible to recover using the old `Aux` trick but that feels like backsliding to Scala 2

core/src/main/scala/coulomb/ops/standard/pow.scala
```diff
@@ -26,18 +26,19 @@ import coulomb.ops.{Pow, SimplifiedUnit}
...
 
-transparent inline given ctx_pow_FractionalPower[V, U, E](using
+given ctx_pow_FractionalPower[V, U, E](using
     alg: FractionalPower[V],
+    dbv: typeexpr.DoubleValue[E],
     su: SimplifiedUnit[U ^ E]
         ): Pow[V, U, E] =
-    val e = typeexpr.double[E]
+    val e = dbv.value
     new Pow[V, U, E]:
```
armanbilge, line 35: Question: instead of making this an anonymous class within the `inline` function, can we make it a named class outside that is then instantiated within the `inline` function? The problem with anonymous classes in `inline` functions is that means everywhere the function is `inline`d, you will get a completely new (anonymous) class. This causes code bloat, esp. on JS.

erikerlandson: I'll play around with it :+1:

erikerlandson: See #265. If I'm understanding what you are thinking, it looks promising

PR #270, opened 2022-04-06
Physical constants

erikerlandson:
@armanbilge   @cquiroz 

A concept for physical constants. The ergonomics looks like this:

```scala
Welcome to Scala 3.1.1 (11.0.14, Java OpenJDK 64-Bit Server VM).
                                                                                                                                            
scala> import coulomb.*, algebra.instances.all.given, coulomb.ops.algebra.all.given, coulomb.policy.standard.given, coulomb.units.si.{*,given}, coulomb.units.si.prefixes.{*,given}, coulomb.units.mksa.{*,given}, coulomb.units.us.{*,given}, coulomb.define.*, coulomb.units.constants.{*,given}
                                                                                                                                            
scala> constq[SpeedOfLight]
val res0: 
  coulomb.quantity.Quantity[coulomb.rational.Rational, coulomb.units.si.Meter / 
    coulomb.units.si.Second
  ] = 299792458
                                                                                                                                            
scala> constq[PlanckConstant]
val res1: 
  coulomb.quantity.Quantity[coulomb.rational.Rational, coulomb.units.mksa.Joule 
    *
   coulomb.units.si.Second] = 132521403/200000000000000000000000000000000000000000
                                                                                                                                            
scala> res0.toValue[Double].show
val res2: String = 2.99792458E8 m/s
                                                                                                                                            
scala> res1.toValue[Double].show
val res3: String = 6.62607015E-34 J s
```

### DIFF
--- a/core/src/main/scala/coulomb/deltaquantity.scala
+++ b/core/src/main/scala/coulomb/deltaquantity.scala
@@ -16,10 +16,10 @@
 
 package coulomb
 
-export deltaquantity.DeltaQuantity
-export deltaquantity.withDeltaUnit
+export dqopaque.DeltaQuantity
+export dqopaque.withDeltaUnit
 
-object deltaquantity:
+object dqopaque:
     import coulomb.ops.*
     import coulomb.conversion.*
 
@@ -103,4 +103,4 @@ object deltaquantity:
         inline def >=[VR, UR](qr: DeltaQuantity[VR, UR, B])(using ord: DeltaOrd[B, VL, UL, VR, UR]): Boolean =
             ord(ql, qr) >= 0
     end extension
-end deltaquantity
+end dqopaque
--- a/core/src/main/scala/coulomb/infra/meta.scala
+++ b/core/src/main/scala/coulomb/infra/meta.scala
@@ -31,6 +31,12 @@ object meta:
             case v if (v == 1) => '{ Rational.const1 }
             case _ => '{ Rational(${Expr(r.n)}, ${Expr(r.d)}) }
 
+    sealed class SigMode
+    object SigMode:
+        case object Canonical extends SigMode
+        case object Standard extends SigMode
+        case object Separate extends SigMode
+
     object rationalTE:
         def unapply(using Quotes)(tr: quotes.reflect.TypeRepr): Option[Rational] =
             import quotes.reflect.*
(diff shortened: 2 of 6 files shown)


erikerlandson:
@armanbilge   I'm not wild about this declaration:

```scala
    // would prefer to use something like `Quantity[Rational, ?]` instead of `Any` here
    transparent inline def constq[CU]: Any = ${ meta.constq[CU] }
```

But `Quantity[Rational, ?]` won't compile, it causes mismatch betwen `[qu]` and `Any`

armanbilge:
Hmm, that's interesting. Does it make a difference in practice though? Since its `transparent inline` I would guess the type is always something more specific anyway.

I think I've seen this pattern before somewhere.

erikerlandson:
@armanbilge   I think in practice `Any` doesn't hurt, because the `transparent` type is always specific, so it just becomes a matter of documenting what it's really returning.  Being able to be more specific, like `Quantity[?, ?]`, would be nice but isn't a blocker.

erikerlandson:
@cquiroz  @armanbilge 
Here's an alternative design that has a proper dependent return type.  It requires an intermediate typeclass, but that's essentially what I'm doing everywhere else:
```scala
scala> import coulomb.*, algebra.instances.all.given, coulomb.ops.algebra.all.given, coulomb.policy.standard.given, coulomb.units.si.{*,given}, coulomb.units.si.prefixes.{*,given}, coulomb.units.mksa.{*,given}, coulomb.units.us.{*,given}, coulomb.define.*, coulomb.units.constants.{*,given}
                                                                                                                                            
scala> constant[Double, SpeedOfLight]
val res0: 
  coulomb.qopaque.Quantity[Double, coulomb.units.si.Meter / 
    coulomb.units.si.Second
  ] = 2.99792458E8
                                                                                                                                            
scala> constant[Double, PlanckConstant].show
val res1: String = 6.62607015E-34 J s
```

this works fairly similar to the way it does in scala-2-coulomb, except accessing actual value is done via single function `constant[V, CU]`, instead of dedicated function for each constant.  Actual constant definitions are all `DerivedUnit`

armanbilge:
Honestly, I think either API is fine from the user-perspective, since the compiler does the right thing anyway :)

It sounds like your alternative design follows a similar pattern to the rest of the lib. That seems nice of its own accord.

PR #272, opened 2022-04-09
spire subproject

erikerlandson:
- coulomb.policy.spire...
- coulomb.conversion.spire.unit
- coulomb.conversion.spire.value
- coulomb.ops.resolution.spire
- coulomb.conversion.coefficients <refactor this?>
- coulomb.ops.algebra.(spire-types...)
- remove coulomb.ops.algebra.all ? (folded into policies)


erikerlandson:
@armanbilge  in previous iterations (scala-2) I have been using `% Provided` on lib dependencies, to avoid having libs show up in my published packages, and allow at least minor flexibility w.r.t. what version a user is already including.  It used to be that anything `% Provided` also showed up in my repl and my unit tests, so it was a clean solution.

Today I tried making `algebra` and `spire` as `% Provided` and it broke both my testing and my repl environment.  I didn't see anything obvious in the latest `sbt` docs, but it seems like something has changed there.

armanbilge:
Hmm, that would be a really unusual thing to do, I rarely rarely see libs use `Provided` since it's pretty bespoke. It's unclear to me what you hope to accomplish with this?

> and allow at least minor flexibility w.r.t. what version a user is already including

A user already has this flexibility; you don't have to do anything.

erikerlandson:
@armanbilge   It was to avoid having the dependencies show up in my published packages.  The first place I ever saw it used was libraries having apache spark as a dep, which is of course quite large.

armanbilge:
The reason it's used for Apache Spark is because Spark is quite literally provided by the runtime, and does not need to be included in the application Jar you deploy. Edit: it would be like shipping with JDK standard libs in your Jar.

As long as a user doesn't depend on the `coulomb-spire` module in their project, then they won't have a spire dependency.

erikerlandson:
> A user already has this flexibility; you don't have to do anything.

How does this work?  If list `spire-0.18.1` as my dep, and someone else imports coulomb, but is already including `spire-0.18.3`, are not both versions of spire now being loaded?

armanbilge:
No, by default* sbt "evicts" the lower-numbered version. This is why MiMa and binary-compatibility is so important. By maintaining backwards-binary-compatibility in spire it guarantees that if I use coulomb with a newer version of spire than you compiled against, everything still works. The entire ecosystem would break down otherwise.

\* with magic incantations, the user can also use an older version if they want.

erikerlandson:
Wild.  well, I don't want to do anything that breaks conventional idioms.   I have been doing this `% Provided` thing on multiple libraries, for years.  OTOH, I have also been NOT ever using MiMa checking :thinking:

armanbilge:
Here, you can actually see eviction at work in coulomb itself :)
```
sbt:root> coreJVM/evicted
[info] Here are other dependency conflicts that were resolved:
[info]  * org.scala-lang:scala3-library_3:3.1.1 is selected over {3.0.2, 3.0.2}
[info]      +- com.manyangled:coulomb-core_3:0.6-6b1b5b7-SNAPSHOT (depends on 3.1.1)
[info]      +- org.typelevel:cats-kernel_3:2.7.0                  (depends on 3.0.2)
[info]      +- org.typelevel:algebra_3:2.7.0                      (depends on 3.0.2)
[success] Total time: 0 s, completed Apr 10, 2022, 1:28:17 AM
```

erikerlandson:
OK then, going forward I will not use Provided, but I will use MiMa. I imagine that will be an adjustment, but it will improve coulomb's ecosystem compliance

armanbilge:
Note that the two concepts are orthogonal to each other: regardless of whether you care about binary-compatibility or not, you should _only_ mark a dependency as provided if it is literally provided by the runtime.

In kind, even though Spark is marked as provided, they _still_ have to care about binary compatibility. If I compile my application against an older version of spark, and submit it to a cluster running a newer version of spark, if spark has not preserved bincompat then my application will crash.

armanbilge:
Sorry to go on this tangent 😂  but here is exactly the sort of problems that arise when spark does break bincompat:
- https://github.com/typelevel/frameless/issues/605

erikerlandson:
Fun fact, my original motivation for `coulomb` was adding units to spark DataSet signatures, but it sort of took on a life of its own :grin:

erikerlandson:
@cquiroz  one interesting thing I noticed was that spire's out-of-box conversions involving BigDecimal use `MathContext.DECIMAL64`:
https://github.com/typelevel/spire/blob/40285e68b035f3ff0d699fb516d005da8bd05d59/core/src/main/scala/spire/math/Convertable.scala#L456

Which means that their representational precision is effectively the same as Double, although intermediate precisions during operations are arbitrary-precision.

I thought I might use DECIMAL128 for storing conversion coefficients:
https://github.com/erikerlandson/coulomb/blob/d89477841f05bed307fa6854fea90c80787b7f8b/spire/src/main/scala/coulomb/conversion/spire/coefficients.scala#L51

cquiroz:
I hadn't notice that.  I think the spire's conversions maybe are going a bit too far deciding that and perhaps should take the precision as a parameter

erikerlandson:
@cquiroz  they support conversions with mathcontext as a parameter, but their implicit conversions all assume decimal64.  Not that this is bad precision.   The only way to "parameterize" the implicit conversions is with another implicit that provides the mathcontext. Which seems possible, but would be another thing you have to import or define.

EDIT: I guess the only thing about this default that seems dubious is that if you're OK with `DECIMAL64` then what's the argument for not just going directly to `Double` ?

erikerlandson:
```
object test:
    import scala.quoted.*
    import scala.language.implicitConversions
 
    abstract class <[L, R]
    
    type A
    type B
    type C
    type D

    type E
    type F
    type G

    given AltB: (A < B) with {}
    given BltC: (B < C) with {}
    given CltD: (C < D) with {}

    given EltF: (E < F) with {}
    given FltG: (F < G) with {}

    given transclose[X, Y, Z](using X < Y, Y < Z): (X < Z) = new (X < Z) {}

    abstract class Resolve[L, R]:
        type O

    transparent inline given XeqX[X]: Resolve[X, X] = new Resolve[X, X] { type O = X }
    transparent inline given LltR[L, R](using L < R): Resolve[L, R] = new Resolve[L, R] { type O = R }
    transparent inline given RltL[L, R](using R < L): Resolve[L, R] = new Resolve[L, R] { type O = L }
```

This feels like cheating but typeclasses can compute transitive closures to create a lattice, and I'm going to use it to clean up `ValueResolution`

erikerlandson:
So tranitive closure only works on total orderings, it fails with ambiguous implicit resolution on any lattice having multiple paths.
Sad!  :joy: 

Two plausible options:
1. auto-generate the code for value resolution pairs algorithmically (aka some flavor of managed build in sbt)
2. write a macro that does a path search.  this is fairly equivalent to asking the implicit resolution system to do a transitive closure but since I'm writing the code it can operate without ambiguous failure.

PR #281, opened 2022-05-13
Add cats instances for Quantity

armanbilge:
These are easy and helpful :)

### DIFF
--- a/core/src/main/scala/coulomb/quantity.scala
+++ b/core/src/main/scala/coulomb/quantity.scala
@@ -16,6 +16,7 @@
 
 package coulomb
 
+import cats.kernel.{Eq, Hash, Order}
 import coulomb.ops.*
 import coulomb.rational.Rational
 import coulomb.conversion.{ValueConversion, UnitConversion}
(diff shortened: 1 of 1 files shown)


core/src/main/scala/coulomb/quantity.scala
```diff
@@ -59,7 +60,15 @@ object qopaque:
...
-            given [U]: Applier[U] = new Applier[U] { def apply[V](v: V): Quantity[V, U] = v } 
+            given [U]: Applier[U] = new Applier[U] { def apply[V](v: V): Quantity[V, U] = v }
+
+        inline given [V, U](using ord: Order[V]): Order[Quantity[V, U]] = ord
+
+    private[qopaque] sealed class QuantityLowPriority0 extends QuantityLowPriority1:
+        inline given [V, U](using hash: Hash[V]): Hash[Quantity[V, U]] = hash
+
+    private[qopaque] sealed class QuantityLowPriority1:
+        inline given [V, U](using eq: Eq[V]): Eq[Quantity[V, U]] = eq
```
armanbilge, line 71: We have to do this annoying prioritisation thing to avoid ambiguous implicits, since both `Order` and `Hash` extends `Eq`.

PR #283, opened 2022-05-14
policy for type aliasing in show

erikerlandson:


armanbilge:
JS tests seem to be getting tripped up by `toString` semantics :)
https://www.scala-js.org/doc/semantics.html

PR #285, opened 2022-05-20
put Quantity extension inside companion object

erikerlandson:


erikerlandson:
cc @armanbilge

erikerlandson:
@armanbilge   I have localized my problem

Here is a reproducer:
```scala

package works {
    opaque type Works[V] = V

    object Works:
        def lift[V](v: V): Works[V] = v

        extension[V](w: Works[V])
            def value: V = w
}

package fails {
    opaque type Fails[V] = V

    object Fails:
        def lift[V](v: V): Fails[V] = v

        extension[V](f: Fails[V])
            def value: V = f

    // adding this makes bad things happen
    extension[V](v: V)
        def asFails: Fails[V] = Fails.lift(v)
}
```

```scala
Welcome to Scala 3.1.2 (11.0.15, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                                                                                
scala> import coulomb.works.*
                                                                                                                                                                                                                
scala> val w = Works.lift(7)
val w: coulomb.works.Works[Int] = 7
                                                                                                                                                                                                                
scala> w.value
val res0: Int = 7
                                                                                                                                                                                                                
scala> import coulomb.fails.*
                                                                                                                                                                                                                
scala> val f = Fails.lift(7)
val f: coulomb.fails.Fails[Int] = 7
                                                                                                                                                                                                                
scala> f.value
val res1: Int = 7

// lifting via the asFails method does NOT work
// notice the type of f2 is also different                                                                                                                                                                                                                
scala> val f2 = 7.asFails
val f2: coulomb.fails.quantity$package.Fails[Int] = 7
                                                                                                                                                                                                                
scala> f2.value
-- [E008] Not Found Error: -----------------------------------------------------
1 |f2.value
  |^^^^^^^^
  |value value is not a member of coulomb.fails.quantity$package.Fails[Int]
1 error found
```

armanbilge:
Wow, nice work! That's weird. Definitely looks like a dotty issue.

What if you put the problematic thing somewhere else? E.g. in a `syntax` package, which is common practice in Typelevel projects anyway.

```
    // adding this makes bad things happen
    extension[V](v: V)
        def asFails: Fails[V] = Fails.lift(v)
```

erikerlandson:
@armanbilge   magic 8-ball says "outlook is positive"

```scala
package works {
    opaque type Works[V] = V

    object Works:
        def lift[V](v: V): Works[V] = v

        extension[V](w: Works[V])
            def value: V = w

    object syntax:
        // defining external lift in subobject works
        extension[V](v: V)
            def asWorks: Works[V] = Works.lift(v)
}
```

```scala
scala> import coulomb.works.*
                                                                                                                                                                                                                
scala> import coulomb.works.syntax.*
                                                                                                                                                                                                                
scala> val w = 7.asWorks
val w: coulomb.works.Works[Int] = 7
                                                                                                                                                                                                                
scala> w.value
val res0: Int = 7
                                                                                                                                                                                                                
scala> val w2 = Works.lift(7)
val w2: coulomb.works.Works[Int] = 7
                                                                                                                                                                                                                
scala> w2.value
val res1: Int = 7
```

erikerlandson:
filed a dotty issue: https://github.com/lampepfl/dotty/issues/15255

erikerlandson:
another successful test, which I believe validates the structure I have in mind
(imagine Fails is Quantity and Works is DeltaQuantity, "both" = "coulomb", etc)

```scala
// interesting fact: these have to be defined in separate files,
// scala seems to frown on defining two `package x {...` in the same source
package both {
    opaque type Fails[V] = V

    object Fails:
        def lift[V](v: V): Fails[V] = v

        extension[V](f: Fails[V])
            def value: V = f

    package syntax {
        export fails.asFails
        object fails:
            extension[V](v: V)
                def asFails: Fails[V] = Fails.lift(v)
    }
}

package both {
    opaque type Works[V] = V

    object Works:
        def lift[V](v: V): Works[V] = v

        extension[V](w: Works[V])
            def value: V = w

    package syntax {
        export works.asWorks
        object works:
            extension[V](v: V)
                def asWorks: Works[V] = Works.lift(v)
    }
}

```

erikerlandson:
@armanbilge  does this new refactoring work with https://github.com/gemini-hlsw/lucuma-core/pull/520 ?

armanbilge:
Thank you so much for refactoring this! It would be helpful for me if you can publish a snapshot from this branch, by adding this to your build and re-generating the workflow:
```scala
ThisBuild / tlCiReleaseBranches += "make-opaque-type-work-somehow"
```

I'll give it a try a bit later!

erikerlandson:
Remembered I need to refactor Hash[Quantity], I will publish a snapshot when I get that

erikerlandson:
@armanbilge  This should work and I don't understand why it isn't:

```scala
Welcome to Scala 3.1.2 (11.0.15, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                                                                                
scala> import coulomb.*, coulomb.syntax.*, algebra.instances.all.given, coulomb.ops.algebra.spire.all.given, coulomb.policy.spire.standard.given, coulomb.units.si.{*,given}, coulomb.units.si.prefixes.{*,given}, coulomb.units.mksa.{*,given}, coulomb.units.us.{*,given}, coulomb.define.*, coulomb.units.constants.{*,given}, coulomb.units.temperature.{*,given}, spire.math.*
                                                                                                                                                                                                                                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Order[Quantity[Int, Meter]]]
val res0: cats.kernel.Order[coulomb.Quantity[Int, coulomb.units.si.Meter]] = coulomb.cats.quantity$infra$QOrder@2d71c745
                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Hash[Quantity[Int, Meter]]]
val res1: cats.kernel.Hash[coulomb.Quantity[Int, coulomb.units.si.Meter]] = coulomb.cats.quantity$infra$QHash@5b635adc
                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Eq[Quantity[Int, Meter]]]
-- Error: ----------------------------------------------------------------------
1 |summon[_root_.cats.kernel.Eq[Quantity[Int, Meter]]]
  |                                                   ^
  |ambiguous implicit arguments: both given instance ctx_Quantity_Order in object standard and given instance ctx_Quantity_Hash in object standard match type cats.kernel.Eq[coulomb.Quantity[Int, coulomb.units.si.Meter]] of parameter x of method summon in object Predef
1 error found
```

armanbilge:
It's very very difficult to disambiguate correctly using `NotGiven`. That's why I opted for the prioritization scheme in https://github.com/erikerlandson/coulomb/pull/281, I believe that should work.

erikerlandson:
@armanbilge   It worries me a little that these hierarchies would depend on the `extends ObjectPriority0...` idiom.  It forces these typeclasses directly into the definition of Quantity, which doesn't seem like good separation of concerns, and doesn't scale well.

I used a similar technique here, to prioritize resolution of a type hierarchy, with success:
https://github.com/erikerlandson/coulomb/blob/scala3/core/src/main/scala/coulomb/ops/standard/pow.scala

erikerlandson:
it didn't even work when I just kept these two rules:
```scala
    given ctx_Quantity_Order[V, U](using ord: Order[V]): Order[Quantity[V, U]] =
        new infra.QOrder[V, U](ord)

    given ctx_Quantity_Hash[V, U](using h: Hash[V]): Hash[Quantity[V, U]] =
        new infra.QHash[V, U](h)
```

which I find even more confusing, since `Hash[Quantity]` ought to satisfy `Eq[Quantity]` at least for any value type satisfying `Hash[V]`, but that also fails.

armanbilge:
This hiearchy is complicated by the fact that `Order` and `Hash` both extend `Eq` independently. Arguably, this was a mistake in Cats and can create all sorts of problems besides this one.

>  It forces these typeclasses directly into the definition of Quantity, which doesn't seem like good separation of concerns, and doesn't scale well.

Typeclasses instances generally have to live in the companion objects anyway, so I don't really see why this is a problem :)

erikerlandson:
> This hiearchy is complicated by the fact that Order and Hash both extend Eq independently.

This explains a lot of my confusion. I must have been looking at obsolete API docs, because in those Order was not a subclass of Eq.

> typeclasses instances generally have to live in the companion objects anyway

I have a philosophical objection to this "have to" part.  It's certainly a common and sensible idiom but I would not want to live in a world where every time someone out in the ecosystem comes up with an interesting new typeclass, I'm obligated to add it directly into `coulomb`. The point-of-integration between a typeclass and some other class in a different package should always be able to be decoupled, IMO.

erikerlandson:
My experiences with hacking around typeclass priorities on `coulomb` have left me with the feeling that an ability to manipulate resolution priorities in some other way besides the `CompanionObject extends CompanionObjectPrio0 ...` idiom would be very very helpful.

armanbilge:
> I have a philosophical objection to this "have to" part.

Sorry, let me qualify that a bit more carefully. What I should have said is:

Typeclass instances generally have to live in the companion objects anyway, **if you want them in scope without imports**.

If you are using imports to bring implicits into scope, then you can import just the right ones so that you don't have ambiguities. Although I agree this is awkward.

I think even the `NotGiven` trick will probably break down in general, these things really only "work" if you make closed-world assumptions.

erikerlandson:
@armanbilge  so it turns out you can use the "prio" trick on a non-companion object:
```scala
object quantity extends quantityprios.quantityPrio1:
    given ctx_Quantity_Order[V, U](using ord: Order[V]): Order[Quantity[V, U]] =
        new QOrder[V, U](ord)

object quantityprios:
    class quantityPrio2:
        class QOrder[V, U](ord: Order[V]) extends Order[Quantity[V, U]]:
            def compare(x: Quantity[V, U], y: Quantity[V, U]) = ord.compare(x.value, y.value)

        class QEq[V, U](e: Eq[V]) extends Eq[Quantity[V, U]]:
            def eqv(x: Quantity[V, U], y: Quantity[V, U]) = e.eqv(x.value, y.value)

        class QHash[V, U](h: Hash[V]) extends QEq[V, U](h) with Hash[Quantity[V, U]]:
            def hash(x: Quantity[V, U]) = h.hash(x.value)

        given ctx_Quantity_Eq[V, U](using e: Eq[V]): Eq[Quantity[V, U]] =
            new QEq[V, U](e)

    class quantityPrio1 extends quantityPrio2:
        given ctx_Quantity_Hash[V, U](using h: Hash[V]): Hash[Quantity[V, U]] =
            new QHash[V, U](h)
```

```scala
Welcome to Scala 3.1.2 (11.0.15, Java OpenJDK 64-Bit Server VM).
                                                                                                                                                                                                                
scala> import coulomb.*, coulomb.syntax.*, coulomb.units.si.{*, given}, coulomb.cats.quantity.given
                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Hash[Quantity[Int, Meter]]]
val res0: cats.kernel.Hash[coulomb.Quantity[Int, coulomb.units.si.Meter]] = coulomb.cats.quantityprios$quantityPrio2$QHash@d6e122
                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Order[Quantity[Int, Meter]]]
val res1: cats.kernel.Order[coulomb.Quantity[Int, coulomb.units.si.Meter]] = coulomb.cats.quantityprios$quantityPrio2$QOrder@4bed2839
                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Eq[Quantity[Int, Meter]]]
val res2: cats.kernel.Order[coulomb.Quantity[Int, coulomb.units.si.Meter]] = coulomb.cats.quantityprios$quantityPrio2$QOrder@4c4da972
```

One annoying thing is that if you try to `export` these, you lose the priority info:
```scala
scala> import coulomb.cats.all.given
                                                                                                                                                                                                                
scala> summon[_root_.cats.kernel.Eq[Quantity[Int, Meter]]]
-- Error: ----------------------------------------------------------------------
1 |summon[_root_.cats.kernel.Eq[Quantity[Int, Meter]]]
  |                                                   ^
  |ambiguous implicit arguments: both given instance ctx_Quantity_Hash in object all and given instance ctx_Quantity_Order in object all match type cats.kernel.Eq[coulomb.Quantity[Int, coulomb.units.si.Meter]] of parameter x of method summon in object Predef
1 error found
```

So I can define these properly exterior to `Quantity` but I don't get to put them in my `policy` definitions.

erikerlandson:
Filed another issue about this export/priority thing: https://github.com/lampepfl/dotty/issues/15264

armanbilge:
Wow, cool, I did not know it works for that as well! :)

FWIW the fact that it doesn't work for `export` is annoying but seems like correct behavior IMO. `export`ing into an object should be equivalent to directly defining aliases on that object which obviously doesn't have any prioritization.

erikerlandson:
@armanbilge agree, I won't be shocked if the conclusion to https://github.com/lampepfl/dotty/issues/15264 is "we won't fix that" :joy:,  OTOH if there's a reasonable way to support it, I think it's the right thing to do.

aside - if they CAN support this, it would mean I could re-write a lot of my operator rules without `NotGiven`

erikerlandson:
@armanbilge  there should be a snapshot to test with now

armanbilge:
Thanks, will give it a try!

PR #289, opened 2022-05-28
Refactor rules to use priorities

erikerlandson:
Priority concept from this issue:
https://github.com/lampepfl/dotty/issues/15264

This works if the `(using PrioX)` is curried as its own separate arglist, so I am going to refactor some of the rules to use priorities instead of `NotGiven`


erikerlandson:
I was hoping `PrioX` could also work for operator rules like `Add`, but the priorities do not fix the problem with ambiguous algebras such as AdditiveSemigroup, where Scala takes possible conversions into account. So far, only the use of `=:=` and `NotGiven[=:=]` allows these rules to work.

PR #307, opened 2022-07-09
support lifted scalar values

### LINKED ISSUE #211: Scalar operations
Should it be allowed to do scalar operations like multiply 2 meters by 4?
I guess addition and substraction are not possible with scalars but multiplication and division should be possible

erikerlandson:
fixes #211

cc @cquiroz  @armanbilge 

Handling left and right hand "raw" (aka scalar) values requires different mechanisms.  This bit of asymmetry is annoying but I don't currently see a way around it. In scala2 I couldn't get it working at all, so it's an improvement.


cquiroz:
Cool, this will be useful
Would it work for other numeric types, let's say multiply by Double or Rational?

erikerlandson:
@cquiroz  it currently defines total support for int, long, float, double.  On the left-hand side I could define all the spire types.  It should already work for any type on the right hand side.

PR #388, opened 2022-11-14
fix and test toInstant with YearsBC

### LINKED ISSUE #383: DeltaUnit for YearsBC not resolving offset correctly?
To experiment with the `DeltaQuantity` feature, I created a DeltaUnit for years BC (obviously only approximate given all the date irregularities in calendars). This differs from the Fahrenheit example in that both the sign and the offset of the unit are negative. When I convert 100BC back into an Instant, it shows as 1870. So. it appears as if the conversion factor, but not the offset, worked as expected.

```
import coulomb.*
import coulomb.syntax.*
import algebra.instances.all.given
import coulomb.ops.algebra.all.given
import coulomb.policy.standard.given
import scala.language.implicitConversions
import coulomb.units.time.{*, given}

object bc:
  final type YearsBC
  given ctx_unit_YearsBC: define.DeltaUnit[YearsBC, -31536000 * Second, -62133800000L, "years BC", "BC"] = define.DeltaUnit()

import bc.*
import coulomb.units.javatime.*
import coulomb.units.javatime.conversions.all.given

@main def testBC =
  println(new java.text.SimpleDateFormat("dd MM yyyy G").parse("01 01 100 BC").toInstant)

  //1870-01-25T00:00:00Z
  println(100.withEpochTime[YearsBC].toInstant)
```

erikerlandson:
fixes #383

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -3,7 +3,7 @@
 // and check in the updates to github workflow yamls
 
 // base version for assessing MIMA
-ThisBuild / tlBaseVersion := "0.6"
+ThisBuild / tlBaseVersion := "0.7"
 
 // publish settings
 ThisBuild / developers += tlGitHubDev("erikerlandson", "Erik Erlandson")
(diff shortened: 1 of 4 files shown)


benhutchison (commented):
The changes in core/src/main/scala/coulomb/conversion/standard/unit.scala look pretty essential, glad my test case helped shake out this feature

PR #392, opened 2022-11-17
refined integration for scala3

erikerlandson:
This initial concept only addresses refined predicates that satisfy algebras such as additive semigroup.

### DIFF
--- a/refined/src/main/scala/coulomb/ops/algebra/refined/all.scala
+++ b/refined/src/main/scala/coulomb/ops/algebra/refined/all.scala
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2022 Erik Erlandson
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package coulomb.ops.algebra.refined
+
+import algebra.ring.AdditiveSemigroup
+
+import eu.timepit.refined.*
+import eu.timepit.refined.api.*
+import eu.timepit.refined.numeric.*
+
+object all:
+    inline given ctx_AdditiveSemigroup_Refined[V](using
+        alg: AdditiveSemigroup[V],
+        vld: Validate[V, Positive]
+    ): AdditiveSemigroup[Refined[V, Positive]] =
+        new AdditiveSemigroup[Refined[V, Positive]]:
+            def plus(x: Refined[V, Positive], y: Refined[V, Positive]): Refined[V, Positive] =
+                refineV[Positive].unsafeFrom(alg.plus(x.value, y.value))
--- a/refined/src/main/scala/coulomb/ops/algebra/refined/all.scala
+++ b/refined/src/main/scala/coulomb/ops/algebra/refined/all.scala
@@ -23,10 +23,20 @@ import eu.timepit.refined.api.*
 import eu.timepit.refined.numeric.*
 
 object all:
-    inline given ctx_AdditiveSemigroup_Refined[V](using
+    inline given ctx_AdditiveSemigroup_Refined_Positive[V](using
         alg: AdditiveSemigroup[V],
         vld: Validate[V, Positive]
     ): AdditiveSemigroup[Refined[V, Positive]] =
         new AdditiveSemigroup[Refined[V, Positive]]:
             def plus(x: Refined[V, Positive], y: Refined[V, Positive]): Refined[V, Positive] =
                 refineV[Positive].unsafeFrom(alg.plus(x.value, y.value))
+
+    inline given ctx_AdditiveSemigroup_Refined_NonNegative[V](using
+        alg: AdditiveSemigroup[V],
+        vld: Validate[V, NonNegative]
+    ): AdditiveSemigroup[Refined[V, NonNegative]] =
+        new AdditiveSemigroup[Refined[V, NonNegative]]:
+            def plus(x: Refined[V, NonNegative], y: Refined[V, NonNegative]): Refined[V, NonNegative] =
+                refineV[NonNegative].unsafeFrom(alg.plus(x.value, y.value))
+
+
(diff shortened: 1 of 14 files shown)


erikerlandson:
xref: https://github.com/erikerlandson/coulomb/discussions/385

refined/src/main/scala/coulomb/ops/algebra/refined/all.scala
```diff
@@ -0,0 +1,32 @@
...
+import eu.timepit.refined.numeric.*
+
+object all:
+    inline given ctx_AdditiveSemigroup_Refined[V](using
+        alg: AdditiveSemigroup[V],
+        vld: Validate[V, Positive]
+    ): AdditiveSemigroup[Refined[V, Positive]] =
+        new AdditiveSemigroup[Refined[V, Positive]]:
+            def plus(x: Refined[V, Positive], y: Refined[V, Positive]): Refined[V, Positive] =
+                refineV[Positive].unsafeFrom(alg.plus(x.value, y.value))
```
armanbilge: what happens if this addition overflows?

erikerlandson: It would be an error, but I haven't worked out a way around it. Technically I think it's type unsound, since if it did overflow it could result in a state not reflected by the type.

erikerlandson: Unless I can check with the validator "manually"

erikerlandson: The only truly type sound semantic would be to have the resulting value be the `Either` expression, which seems awkward to me, however it would work.

erikerlandson: Well, it "works" in the sense that any type can be a Quantity value, but the algebra typeclass requires the original V type.

armanbilge: You might have a look at this:
https://github.com/gemini-hlsw/refined-algebra/blob/97c3510db3912a0dd57679f11a311644af7c61f8/refined-algebra/src/main/scala/refined/algebra.scala#L13

erikerlandson: `unsafeFrom` throws a runtime exception if it overflows, or the result otherwise violates the predicate.  I suppose it is a matter of philosophy whether it's "better" to error out, or not error out but silently return an incorrect result (i.e. max-int).  Neither is great, although my intuition is that if I'm getting overflows, I'd want to know it via a runtime error, instead of getting a result that doesn't reflect the actual sum.
 
```scala
scala> refineV[Positive].unsafeFrom(2000000000)
val res3: Int Refined eu.timepit.refined.numeric.Positive = 2000000000
                                                                                                                                                           
scala> refineV[Positive].unsafeFrom(2000000000 + 2000000000)
java.lang.IllegalArgumentException: Predicate failed: (-294967296 > 0).
  at eu.timepit.refined.internal.RefinePartiallyApplied.unsafeFrom$$anonfun$1(RefinePartiallyApplied.scala:21)
  at scala.util.Either.fold(Either.scala:190)
  at eu.timepit.refined.internal.RefinePartiallyApplied.unsafeFrom(RefinePartiallyApplied.scala:21)
  ... 66 elided
```

erikerlandson: At any rate, the current behavior is NOT unsound, at least in the sense that it won't silently return a Refined with a value that violates the predicate. It will either return a correct value or error out at runtime.

armanbilge: > At any rate, the current behavior is NOT unsound

But note that it is unlawful.

erikerlandson: Also probably depends on whether you are semantically trying to represent a modular ring (as in that example) or an unbounded additive semigroup.

erikerlandson: On the bright side, `AdditiveSemigroup` for `+`  will admit either my algebra definitions or a user's preferred definition, whichever they want to import.

erikerlandson: FWIW, one of my design philosophies has been that the "out-of-box" definitions provided by coulomb deliberately do not try to obscure the "native" behaviors of numeric types.  So integer addition can overflow.  Floating points do whatever complicated native floating point behaviors, etc.

Users may very reasonably wish to define better semantics, but the definitions of "better" have a lot of potential variation, and I decided it was clarifying to me to avoid that kind of potential for policy explosion.

armanbilge: Sure. FWIW, the design philosophy of algebra `AdditiveSemigroup` and friends is that implementing instances are lawful. The laws _are_ the semantics.

erikerlandson:
the "algebraic" policy overlays seem to work:

```scala
scala> import eu.timepit.refined.*, eu.timepit.refined.api.*, eu.timepit.refined.numeric.*, coulomb.policy.standard.{*, given}, algebra.ring.*, algebra.instances.all.given, coulomb.units.si.{*, given}, coulomb.units.si.prefixes.{*, given}, coulomb.*, coulomb.syntax.*, coulomb.policy.overlay.refined.algebraic.given
                                                                                                                                
scala> refineV[Positive].unsafeFrom(1d).withUnit[Meter]
val res0: 
  coulomb.Quantity[Double Refined eu.timepit.refined.numeric.Positive, 
    coulomb.units.si.Meter
  ] = 1.0
                                                                                                                                
scala> refineV[Positive].unsafeFrom(1).withUnit[Kilo * Meter]
val res1: 
  coulomb.Quantity[Int Refined eu.timepit.refined.numeric.Positive, 
    coulomb.units.si.prefixes.Kilo
   * coulomb.units.si.Meter] = 1
                                                                                                                                
scala> res0 + res1
val res2: 
  coulomb.Quantity[Double Refined 
    eu.timepit.refined.numeric.Greater[shapeless.nat._0]
  , coulomb.units.si.Meter] = 1001.0
```

benhutchison:
Looks very promising. Just yesterday was experimenting with porting some code onto Coulomb, where I divided 24 hours by a short time interval duration, to calculate how many such intervals occur per day. And it made me think that once this refinement is available (for division), the result could be inferred as positive. Currently, I re-assert the result is positive since I don't have inference..

erikerlandson:
xref https://github.com/fthomas/refined/issues/932

benhutchison:
@erikerlandson If you're willing to cut a beta or RC binary release off this branch sometime, I'd be keen to experiment with it in an application codebase and provide feedback.

benhutchison:
I'm a little unsure if you're dependent on changes in [Refined](https://github.com/fthomas/refined/issues/932) to make this viable or if you consider it near usable in present form.

erikerlandson:
@benhutchison  I cut a prerelease 0.7.1-RC1
https://github.com/erikerlandson/coulomb/releases/tag/v0.7.1-RC1

erikerlandson:
@benhutchison regarding https://github.com/fthomas/refined/issues/932, I am not blocked on that. It basically means (for now) I will be providing `refineVU[P, U]` but NOT `refineMVU[P, U]` (the compile-time checking version)

PR #397, opened 2022-11-20
I can't explain why this doesnt compile?

benhutchison:


benhutchison:
I was seeing the problem in my own code and went looking for the unit test that proved it works. This seemed to be the right place but it didnt work here either... confused..

erikerlandson:
it is because your value is an integer, which means the result will (or may) lose precision from integer truncation.

coulomb distinguishes between value and unit conversions, and "truncating" value and unit conversions.  the truncating methods are: `q.tToValue` and `q.tToUnit`.  These are used when your value types are integral (int, long, bigint, etc).  Additionally, truncating conversions are *only* explicit: you have to explicitly invoke `tToValue` or `tToUnit`.

erikerlandson:
essentially, these are the analog of the difference between `/` and `quot` , or `multiplicative-group' and 'truncatingdivision' in the typelevel algebra definitions.

erikerlandson:
you can see examples of `tToUnit` and friends in the unit tests - I also have tests which verify that the non-truncating versions fail on integral types

benhutchison:
Thanks for the explanation Erik

Issue #405, opened 2022-12-02
2 Issues:  Simplifying and exporting time units


benhutchison:
1.

```scala
case class TimeUnit(code: String, interval: Quantity[Long, coulomb.units.time.Second]):
  def intervalsPerDay: Quantity[Long, 1] = 24L.withUnit[Hour].tquot(interval)
//  def intervalsPerDay: Quantity[Long, coulomb.units.time.Hour / coulomb.units.si.Second] = 24L.withUnit[Hour].tquot(interval)
```

Surprised it doesnt compile with:
```
Found:    coulomb.Quantity[?1.VO, ?1.UO]
Required: coulomb.Quantity[Long, (1 : Int)]

where:    ?1 is an unknown value of type coulomb.ops.standard.tquot.infra.TQuotNC[Long, coulomb.units.time.Hour, Long, 
  coulomb.units.si.Second
, Long, coulomb.units.time.Hour / coulomb.units.si.Second]

  def intervalsPerDay: Quantity[Long, 1] = 24L.withUnit[Hour].tquot(interval)
```

If I uncomment the unsimplified unit `coulomb.units.time.Hour / coulomb.units.si.Second` it compiles, but of course I'd like hours and seconds to cancel out to dimensionless.

Could there be an issue with simplification of time unit types?

2.

Separately, Ive been experimenting with _exporting_ all the required Coulomb definitions. I have a file `Basetypes.scala` that builds on top of Coulomb, and all my other files import is like so `import basetypes.{*, given}`.

Within `Basetypes.scala` I export Coulomb to try to centralize all the imports and not duplicates them across 10s of files. It seems to be working, but note I had to hide the re-export of `Second` from `si` into `time` to avoid a double-definition error.

```
package basetypes

export coulomb.{Quantity, DeltaQuantity, `*`, `/`, `^`, showUnit}
export coulomb.syntax.{withUnit, withDeltaUnit}
export allalgebra.given
export coulomb.ops.algebra.all.given
export coulomb.policy.standard.given
export coulomb.units.time.{*, given}
export coulomb.units.si.{Second as _, ctx_unit_Second as _, *, given}
export coulomb.units.si.prefixes.{*, given}

private object allalgebra extends algebra.instances.AllInstances
```

To be clear this certainly isn't a _bug_, but it is a potential issue that users of Coulomb may encounter. This need to hide re-exported members seems a small issue now, but it grows larger & problematic the more exporting is utilized. 

[...]

PR #406, opened 2022-12-02
Expect this to compile as time units should cancel to be unitless?

benhutchison:
ref https://github.com/erikerlandson/coulomb/issues/405

### DIFF
--- a/units/src/test/scala/coulomb/time.scala
+++ b/units/src/test/scala/coulomb/time.scala
@@ -68,6 +68,11 @@ class TimeUnitsSuite extends CoulombSuite:
         import coulomb.policy.standard.given
 
         36d.withEpochTime[Hour].toUnit[Day].assertDQD[Double, Day](1.5)
+        
+        //I think this ought to compile?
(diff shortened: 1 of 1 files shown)


benhutchison:
```
        val s = summon[SimplifiedUnit[Minute / Second]]
        println(summon[ShowUnit[s.UO]].value)  // min/s
```

erikerlandson:
in theory, that unit test could compile via implicit `Conversion` - however I deliberately did not supply implicit conversions for truncating conversions.

It would be possible to provide those. If I did, I would make it some kind of deliberate import, like `import coulomb.conversion.standard.truncating.given` perhaps.

I still don't think it's a safe idea - it works OK when the integer divisions come out even but for example if you inverted it, the result would truncate to zero.

benhutchison:
Right, makes sense. It's not safe in general, but such an implicit import might be an opt-in that, in a particular context, the conversion is considered safe.

If after more usage, I'm convinced such an implicit adds value, I'll be sure to mention it..

PR #410, opened 2022-12-15
Non-compiling example

benhutchison:
Showing subtraction problem between DeltaQty and Qty when a) the units are commensurate but not identical and b) exported unit type aliases are involved.

### DIFF
--- a/units/src/test/scala/coulomb/time.scala
+++ b/units/src/test/scala/coulomb/time.scala
@@ -143,6 +143,9 @@ class TimeUnitsSuite extends CoulombSuite:
         // both different
         (61L.withEpochTime[Second] - 1f.withUnit[Minute])
             .assertDQ[Float, Second](1)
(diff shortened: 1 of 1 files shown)


benhutchison:
Actually, the error msg here was `Subtraction not defined in scope for DeltaQuantity[Long, coulomb.units.si.prefixes.Milli * coulomb.units.time.Second] and Quantity[Long, coulomb.units.si.Second]`

This led me to believe the problem related to `export` of `Second`. But this doesnt seem to be the case.

The real problem is simply that subtraction of `Second` from `Milli*Second` isn't allowed. 

I had thought this would work if the units are [Commensurable](https://erikerlandson.github.io/coulomb/concepts.html#unit-conversions). Ie that the result would be in the more fine-grained unit of the operands. No such luck, they need be aligned prior to doing arithmetic.

PR #417, opened 2023-01-09
Add refined module to root

armanbilge:
Unfortunately it did not publish, because it was not added to the root aggregate. Sorry!

Edit: also, because it was not added to the root, it was not being checked in CI, including formatting, MiMa, tests, etc.


erikerlandson:
@armanbilge  thanks, sorry I didn't catch that!

PR #526, opened 2023-11-05
A concept of abstract physical quantities

erikerlandson:
An experimental pull request based on this discussion:
https://github.com/erikerlandson/coulomb/discussions/525


erikerlandson:
I'm going to close this for now, if I come up with better ways to think about this that make sense for coulomb I can re-open it

Issue #50, opened 2019-09-23
sequences / collections with units


erikerlandson:
CoulombVector, CoulombList, etc.  It is straightforward to declare Vector[Quantity[...]]  but more efficient to abstract out the unit types

erikerlandson:
for scala 3 collection design and implementations
- https://docs.scala-lang.org/overviews/core/custom-collections.html
- https://docs.scala-lang.org/overviews/core/architecture-of-scala-213-collections.html

PR #530, opened 2023-11-10
Coulomb collections

erikerlandson:
An attempt to implement #50

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -80,11 +80,11 @@ jobs:
 
       - name: Make target directories
         if: github.event_name != 'pull_request' && (startsWith(github.ref, 'refs/tags/v') || github.ref == 'refs/heads/scala3')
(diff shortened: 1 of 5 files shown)


erikerlandson:
As an efficiency measure I tried creating a subclass of `scala.Conversion` that can also operate directly on "raw" values inside a QuantityVector.  Doing this will break ABI compatibility, and so I wanted to verify whether it actually improves efficiency before I commit to it.

I benchmarked this using a couple of test functions:

```scala
def bench1[V, U](qv: QuantityVector[V, U])(using cnv: scala.Conversion[Quantity[V, U], Quantity[Double, Meter]]): QuantityVector[Double, Meter] =
    qv.map(cnv)

def bench2[V, U](qv: QuantityVector[V, U])(using cnv: QuantityConversion[V, U, Double, Meter]): QuantityVector[Double, Meter] =
    QuantityVector[Meter](qv.values.map(cnv.raw))
```

**The `bench2` variation that uses the raw conversion is > 7x faster.**  Furthermore, it produces significantly less load on the garbage collection, so the benefits are even bigger.

This demonstrates the benefits of breaking MIMA with `QuantityConversion`, and also justifies the benefits of defining `QuantityVector` in the first place.

erikerlandson:
merge conflicts after 0.9.0 are such that I am just going to create a new PR starting from 0.90

